Issue #9, opened 2019-01-15
Error during WebSocket handshake: 'Upgrade' header is missing


tonylin0826:
When using chrome as websocket client, it failed with the following message. 

![image](https://user-images.githubusercontent.com/5596402/51165746-3bffc380-18dc-11e9-9ebb-51553916f3e5.png)

Seems like the server didn't send the correct header.

bsergean:
It should be good now that to your fix !

I added fdbd213fa28cf9b1420f9c96c33342af115cfa2e to properly handle the Connection header. Tell me if this is giving you trouble with Chrome.

PR #10, opened 2019-01-15
Fix missing "Upgrade" header error

tonylin0826:
Hi, this is a quick fix of this problem. I've create a new [Issue](https://github.com/machinezone/IXWebSocket/issues/9#issue-399222657) at your repo. Hope it helps.


tonylin0826:
Great!

You're right, I put the wrong value in connection flag, it should be "Upgrade".

PR #27, opened 2019-04-15
Ping always sent, option to disconnect on no pong response option, close code fix

Kumamon38:
Hi,
I wanted to have a regular ping no matter what we send or receive, so I changed a bit how it was done and created a thread for that.
I also added an option to disconnect if no PONG are received for a given time (factor of heartBeatPeriod) : `ixWebsocket.enableDisconnectOnNoHeartBeatResponse(2);` to wait for 2 heartbeat period 
And I fixed an issue with the close code in case it is received in IXWebSocketTransport.cpp, `close()` method
```
        if(_closeCode == 0)
        {
            _closeCode = 1000;
            _closeReason = "Normal Closure";
        }
```

Feel free to modify it as you want :)

### DIFF
--- a/ixwebsocket/IXWebSocketTransport.cpp
+++ b/ixwebsocket/IXWebSocketTransport.cpp
@@ -75,7 +75,7 @@ namespace ix
 
     WebSocketTransport::~WebSocketTransport()
     {
-        stopHeartBeat();
+        stopHeartBeatWatchdogThread();
     }
 
     void WebSocketTransport::configure(const WebSocketPerMessageDeflateOptions& perMessageDeflateOptions,
@@ -86,13 +86,13 @@ namespace ix
         _heartBeatPeriod = heartBeatPeriod;
         _heartBeatFactorDisconnectOnNoResponse = heartBeatFactorDisconnectOnNoResponse;
 
-        stopHeartBeat();
+        stopHeartBeatWatchdogThread();
 
         if(_heartBeatPeriod != kDefaultHeartBeatPeriod)
         {
             _exitSignal = std::promise<void>();
             _future = _exitSignal.get_future();
-            _thread = std::thread(&WebSocketTransport::runHeartBeat, this);
+            _heartbeatWatchdogThread = std::thread(&WebSocketTransport::heartbeatWatchdogThreadFunc, this);
         }
     }
 
@@ -199,7 +199,7 @@ namespace ix
 
     bool WebSocketTransport::pongReceiveDelayExceeded()
     {
-        if (_heartBeatFactorDisconnectOnNoResponse == kDefaultHeartBeatFactorDisconnectOnNoResponse)
+        if (_heartBeatFactorDisconnectOnNoResponse <= 0)
             return false;
 
         std::lock_guard<std::mutex> lock(_lastReceivePongTimePointMutex);
@@ -207,18 +207,16 @@ namespace ix
         return now - _lastReceivePongTimePoint > std::chrono::seconds(_heartBeatFactorDisconnectOnNoResponse*_heartBeatPeriod);
     }
 
-    void WebSocketTransport::runHeartBeat()
+    void WebSocketTransport::heartbeatWatchdogThreadFunc()
     {
-        using millis = std::chrono::duration<double, std::milli>;
-
-        while (_future.wait_for(std::chrono::milliseconds(0)) == std::future_status::timeout)
+        while (_future.wait_for(std::chrono::milliseconds(1000*_heartBeatPeriod)) == std::future_status::timeout)
         {
             if(_readyState == OPEN)
             {
                 // if (1) disconnect on no PONG received is enabled and (2) heartbeat is enabled and (3) duration exceeded the maximum delay, then close the connection
                 if(pongReceiveDelayExceeded())
                 {
-                    close();
+                    close(4000, "Heartbeat no answer failure");
                 }
                 // If (1) heartbeat is enabled and (2) send for a duration exceeding our heart-beat period, send a ping to the server.
                 else if (heartBeatPeriodExceeded())
@@ -228,17 +226,15 @@ namespace ix
                     sendPing(ss.str());
                 }
             }
-            
-            std::this_thread::sleep_for(millis(1000*_heartBeatPeriod));
         }
     }
     
-    void WebSocketTransport::stopHeartBeat()
+    void WebSocketTransport::stopHeartBeatWatchdogThread()
     {
-        if (_thread.joinable()) // we've already been started
+        if (_heartbeatWatchdogThread.joinable()) // heart beat is running
         {
             _exitSignal.set_value();
-            _thread.join();
+            _heartbeatWatchdogThread.join();
         }
     }
 
@@ -523,14 +519,7 @@ namespace ix
                 std::string reason(_rxbuf.begin()+ws.header_size + 2,
                                    _rxbuf.begin()+ws.header_size + 2 + (size_t) ws.N);
 
-                {
-                    std::lock_guard<std::mutex> lock(_closeDataMutex);
-                    _closeCode = code;
-                    _closeReason = reason;
-                    _closeWireSize = _rxbuf.size();
-                }
-
-                close();
+                close(code, reason, _rxbuf.size());
             }
             else
             {
@@ -782,8 +771,11 @@ namespace ix
 
         WebSocketSendInfo info = sendData(wsheader_type::PING, message, compress);
 
-        std::lock_guard<std::mutex> lck(_lastSendPingTimePointMutex);
-        _lastSendPingTimePoint = std::chrono::steady_clock::now();
+        if(info.success)
+        {
+            std::lock_guard<std::mutex> lck(_lastSendPingTimePointMutex);
+            _lastSendPingTimePoint = std::chrono::steady_clock::now();
+        }
 
         return info;
     }
@@ -833,7 +825,7 @@ namespace ix
         }
     }
 
-    void WebSocketTransport::close()
+    void WebSocketTransport::close(uint16_t code, const std::string& reason, size_t closeWireSize)
     {
         _requestInitCancellation = true;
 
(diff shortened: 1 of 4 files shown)


ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -730,7 +779,13 @@ namespace ix
     WebSocketSendInfo WebSocketTransport::sendPing(const std::string& message)
     {
         bool compress = false;
-        return sendData(wsheader_type::PING, message, compress);
+
+        WebSocketSendInfo info = sendData(wsheader_type::PING, message, compress);
+
+        std::lock_guard<std::mutex> lck(_lastSendPingTimePointMutex);
+        _lastSendPingTimePoint = std::chrono::steady_clock::now();
```
Dimon4eg: I think we should save `_lastSendPingTimePoint` only if send was successful

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -176,33 +189,67 @@ namespace ix
...
+        std::lock_guard<std::mutex> lock(_lastReceivePongTimePointMutex);
+        auto now = std::chrono::steady_clock::now();
+        return now - _lastReceivePongTimePoint > std::chrono::seconds(_heartBeatFactorDisconnectOnNoResponse*_heartBeatPeriod);
+    }
+
+    void WebSocketTransport::runHeartBeat()
+    {
+        using millis = std::chrono::duration<double, std::milli>;
+
+        while (_future.wait_for(std::chrono::milliseconds(0)) == std::future_status::timeout)
```
Dimon4eg: Better to wait with timeout `millis(1000*_heartBeatPeriod)` and remove `sleep_for` below

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -176,33 +189,67 @@ namespace ix
...
-        return now - _lastSendTimePoint > std::chrono::seconds(_heartBeatPeriod);
+        return now - _lastSendPingTimePoint > std::chrono::seconds(_heartBeatPeriod);
     }
 
-    void WebSocketTransport::poll()
+    bool WebSocketTransport::pongReceiveDelayExceeded()
     {
-        _socket->poll(
-            [this](PollResultType pollResult)
+        if (_heartBeatFactorDisconnectOnNoResponse == kDefaultHeartBeatFactorDisconnectOnNoResponse)
```
Dimon4eg: If `kDefaultHeartBeatFactorDisconnectOnNoResponse` will be changed in future to 2 for example, this code will not work

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -176,33 +189,67 @@ namespace ix
...
                 }
+            }
+            
+            std::this_thread::sleep_for(millis(1000*_heartBeatPeriod));
+        }
+    }
+    
+    void WebSocketTransport::stopHeartBeat()
+    {
+        if (_thread.joinable()) // we've already been started
```
Dimon4eg: comment copied from `start` method :)

ixwebsocket/IXWebSocketTransport.h
```diff
@@ -155,14 +157,25 @@ namespace ix
         std::atomic<bool> _requestInitCancellation;
 
         // Optional Heartbeat
+        std::thread _thread;
```
Dimon4eg: better to rename `_thread` to for example `_heartbeatWatchdogThread`

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -176,33 +189,67 @@ namespace ix
...
-            [this](PollResultType pollResult)
+        if (_heartBeatFactorDisconnectOnNoResponse == kDefaultHeartBeatFactorDisconnectOnNoResponse)
+            return false;
+
+        std::lock_guard<std::mutex> lock(_lastReceivePongTimePointMutex);
+        auto now = std::chrono::steady_clock::now();
+        return now - _lastReceivePongTimePoint > std::chrono::seconds(_heartBeatFactorDisconnectOnNoResponse*_heartBeatPeriod);
+    }
+
+    void WebSocketTransport::runHeartBeat()
```
Dimon4eg: Better name it `heartbeatWatchdogThreadFunc`

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -176,33 +189,67 @@ namespace ix
...
             {
-                // If (1) heartbeat is enabled, and (2) no data was received or
-                // send for a duration exceeding our heart-beat period, send a
-                // ping to the server.
-                if (pollResult == PollResultType::Timeout &&
-                    heartBeatPeriodExceeded())
+                // if (1) disconnect on no PONG received is enabled and (2) heartbeat is enabled and (3) duration exceeded the maximum delay, then close the connection
+                if(pongReceiveDelayExceeded())
+                {
+                    close();
```
Dimon4eg: I guess we need to close with some reason, maybe 1006 - Abnormal Closure.

ixwebsocket/IXWebSocketTransport.cpp
Kumamon38: Agree

ixwebsocket/IXWebSocketTransport.cpp
Kumamon38: I think we won't change `kDefaultHeartBeatFactorDisconnectOnNoResponse`because it is -1 to disable by default

ixwebsocket/IXWebSocketTransport.h
Kumamon38: agree

ixwebsocket/IXWebSocketTransport.cpp
Kumamon38: agree

ixwebsocket/IXWebSocketTransport.cpp
Kumamon38: 1006 seems to be for truncated payload and things like that, so I'm not sure it is for what we have done here

ixwebsocket/IXWebSocketTransport.cpp
Dimon4eg: currently yes but let's imagine in future we will enable it bu default

ixwebsocket/IXWebSocketTransport.cpp
Kumamon38: agree

ixwebsocket/IXWebSocketTransport.cpp
Dimon4eg: ok, use some another code and not 1000.
Also set some reason to _closeReason

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -776,34 +823,32 @@ namespace ix
...
-        // We use 1000: normal closure.
-        //
-        // >>> struct.pack('!H', 1000)
-        // b'\x03\xe8'
-        //
-        const std::string normalClosure = std::string("\x03\xe8");
+        
+        std::stringstream stream;
+        stream << std::hex << code;
+        const std::string closure(stream.str());
```
Dimon4eg: this is not correct. 
For example for code 1000 it will send text "3e8" instead of "\x03\xe8"

bsergean:
Hey thanks @Kumamon38 for the pull request and @Dimon4eg for looking at it in details. Some thoughts.

1. I like the idea of auto-closing when the remote endpoint does not respond. I am using a [python library for servers](https://websockets.readthedocs.io/en/stable/intro.html) and it has this feature. However I am not in love of needing a new thread to implement it. I'd like to think about a way to implement it without a thread, by reusing the existing timeout from select. If you can think about that and propose something please go ahead.
2. The close code isn't great with this hard-coded value. I've been meaning to clean that up, but also to allow passing an arbitrary value when closing. So we'd have to expose this to the WebSocket class. Maybe having a separate PL or commit just for that would be best.
3. It would be best to have unittest for those new modes. Ideally we'd also test against other servers, and have that automated (more work but not impossible).
4. That's more of a indent nitpick but I put spaces after my if / for statements. I should get a clang-format script in place or something like that.

ixwebsocket/IXWebSocketTransport.cpp
Kumamon38: Right, I’ll change that

Kumamon38:
No problem @bsergean  :)
1. I'll look at it without thread
2. I submitted a PR just with that
3. It would be better yeah ;)
4. I'll try to remind that

Please delete this PR

Kumamon38:
@bsergean I have submitted the PR with code, and I'm working on the 1. and 3. points currently, but I can't find a way to make my auto disconnect test with what we have currently. To test we would need to prevent the server to send a PONG back, or prevent the client to read it. As the server uses the same class WebSocket, the only way to do that would be to disable the auto PONG answers by adding a method, but adding this method just for the test is not very good. What do you think about it?

bsergean:
Yes you're right that it's a bit odd to add such a functionality. One thing that is used by drivers is environment variables, it's a bit ugly too but isn't as intrusive API wise.

Something we could do to is to change the way the callback system work, a bit like HTML where we can 'accept' events. So from that callback we would have a way to 'cancel' some events. But that sounds complicated.

I think that my preference goes with having WebSocketTransport::configure take an extra bool 'respondToPing' that default to true. And also WebSocket have something similar. What would be good would be to add such an option to the command line tool ws, say for the echo_server class. And then use that as an interactive way to make sure that the heartbeat works as expected.

```
        void enableRespondToPing();
        void disableRespondToPing();
```

bsergean:
If you come up with a better idea for a name that RespondToPing, please go ahead, I'm not convinced it's the best name.

Kumamon38:
Thanks for the feedback, I’ll see that tomorrow (not sure to add this in every place you mentioned for now). The PR is already ready without these needed features, so might not take too long. As for the name, maybe `enablePong` could work too :)

bsergean:
Cool. 

For the whole heartbeat feature, I have just checked again the [documentation of websockets](https://websockets.readthedocs.io/en/stable/api.html#websockets.protocol.WebSocketCommonProtocol).

```
Once the connection is open, a Ping frame is sent every ping_interval seconds. This serves as a keepalive. It helps keeping the connection open, especially in the presence of proxies with short timeouts. Set ping_interval to None to disable this behavior.

If the corresponding Pong frame isn’t received within ping_timeout seconds, the connection is considered unusable and is closed with status code 1011. This ensures that the remote endpoint remains responsive. Set ping_timeout to None to disable this behavior.
```

I think that those would be good names: **setPingInterval** (would be an alias for setHeartBeatPeriod), and **setPingTimeout** would be what your original PR was about. We wouldn't define that timeout in term of a factor of the hearbeatperiod, but instead we'd define it as an absolute value (in seconds). If it is -1, then that feature isn't activated.

Kumamon38:
Yes, and so change my current code from 4000 to 1011

PR #28, opened 2019-04-16
close method change and fix code

Kumamon38:
Separate changes from previous PR

### DIFF
--- a/ixwebsocket/IXWebSocketTransport.cpp
+++ b/ixwebsocket/IXWebSocketTransport.cpp
@@ -474,14 +474,8 @@ namespace ix
                 std::string reason(_rxbuf.begin()+ws.header_size + 2,
                                    _rxbuf.begin()+ws.header_size + 2 + (size_t) ws.N);
 
-                {
-                    std::lock_guard<std::mutex> lock(_closeDataMutex);
-                    _closeCode = code;
-                    _closeReason = reason;
-                    _closeWireSize = _rxbuf.size();
-                }
 
-                close();
+                close(code, reason, _rxbuf.size());
             }
             else
             {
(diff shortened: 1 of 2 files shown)


ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -781,29 +775,32 @@ namespace ix
         _lastSendTimePoint = std::chrono::steady_clock::now();
     }
 
-    void WebSocketTransport::close()
+    void WebSocketTransport::close(uint16_t code, const std::string& reason)
```
Dimon4eg: instead of two `close` methods you can add default value 0 to `closeWireSize` to original `close`

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -781,29 +775,32 @@ namespace ix
...
-        sendData(wsheader_type::CLOSE, normalClosure, compress);
+        sendData(wsheader_type::CLOSE, closure, compress);
         setReadyState(CLOSING);
 
         _socket->wakeUpFromPoll(Socket::kCloseRequest);
         _socket->close();
 
-        _closeCode = 1000;
-        _closeReason = "Normal Closure";
+        _closeCode = code;
```
Dimon4eg: probably you need to protect data with `_closeDataMutex`

Dimon4eg:
Looks fine :+1:

ixwebsocket/IXWebSocketTransport.cpp
Kumamon38: Added

bsergean:
LGTM. Nice to have will be to add extra tests, and add the extra parameters to the WebSocket class when closing.

bsergean:
FYI just did that bcf2fc18129dddb2d76095705d1a8afeb4869688

PR #29, opened 2019-04-16
fix windows build

Dimon4eg:
It builds but have not tested it yet :)

### DIFF
--- a/ws/CMakeLists.txt
+++ b/ws/CMakeLists.txt
@@ -7,7 +7,9 @@ cmake_minimum_required (VERSION 3.4.1)
 project (ws)
 
 # There's -Weverything too for clang
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic")
+if (NOT WIN32)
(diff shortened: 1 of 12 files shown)


Dimon4eg:
@bsergean can we add windows build to CI?

ws/CMakeLists.txt
```diff
@@ -51,4 +53,13 @@ add_executable(ws
   ws.cpp)
 
 target_link_libraries(ws ixwebsocket)
+
```
bsergean: Does that one means that we need OpenSSL on Windows ?
Could we make that one use the TLS macro, or be Linux only ?

bsergean:
Yes it would be good to have Windows setup on CI, but Windows platform wasn't available on travis when I started this.

https://blog.travis-ci.com/2018-10-11-windows-early-release
https://docs.travis-ci.com/user/reference/windows/

In the forums they explain how to setup msvc by running some batch files.

ws/CMakeLists.txt
Dimon4eg: we don't build we on windows currently as there are many errors.

PR #32, opened 2019-04-17
Real ping

Kumamon38:
- Added real ping
- Added Ping timeout
- Rename few methods
- Added tests


Kumamon38:
@bsergean So here is the PR, I added test cases and changed what we discussed yesterday, did some renaming :)
I didn't add the `enablePong()` in `ws` for now

ixwebsocket/IXWebSocket.h
```diff
@@ -152,9 +158,15 @@ namespace ix
         std::atomic<int> _handshakeTimeoutSecs;
         static const int kDefaultHandShakeTimeoutSecs;
 
-        // Optional Heartbeat
-        int _heartBeatPeriod;
-        static const int kDefaultHeartBeatPeriod;
+        // enable or disable PONG frame response to received PING frame
+        bool _enablePong;
+        static const bool kDefaultEnablePong;
```
Dimon4eg: I prefer to assign constant values directly in header.

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -51,10 +51,22 @@
 #include <thread>
 
 
+int gcd (int a, int b) {
```
Dimon4eg: Greatest common divisor?
Rename it to clear name or add at least comment

ixwebsocket/IXWebSocket.h
Kumamon38: @bsergean is always doing that in the body, so I keep the same way

ixwebsocket/IXWebSocketTransport.cpp
Kumamon38: Renamed

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -51,10 +51,22 @@
 #include <thread>
```
bsergean: indent nitpick / we put curly braces on the next line as a convention in our code base. I'll fix that myself.

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -176,33 +202,51 @@ namespace ix
...
+        std::lock_guard<std::mutex> lock(_lastReceivePongTimePointMutex);
+        auto now = std::chrono::steady_clock::now();
+        return now - _lastReceivePongTimePoint > std::chrono::seconds(_pingTimeoutSecs);
     }
 
     void WebSocketTransport::poll()
     {
+        if (_readyState == OPEN)
+        {
+            // if (1) ping tiemout is enabled and (2) duration exceeded the maximum delay, then close the connection
```
bsergean: typo: tiemout -> timeout

ixwebsocket/IXWebSocketTransport.h
```diff
@@ -68,7 +68,8 @@ namespace ix
         ~WebSocketTransport();
 
         void configure(const WebSocketPerMessageDeflateOptions& perMessageDeflateOptions,
```
bsergean: the indent nerd in me sees extra spaces here.

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -444,12 +488,16 @@ namespace ix
             else if (ws.opcode == wsheader_type::PING)
             {
                 unmaskReceiveBuffer(ws);
+
                 std::string pingData(_rxbuf.begin()+ws.header_size,
-                                     _rxbuf.begin()+ws.header_size + (size_t) ws.N);
+                                         _rxbuf.begin()+ws.header_size + (size_t) ws.N);
```
bsergean: Was that line not indented properly ? New indent doesn't look correct on github.

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -176,33 +202,51 @@ namespace ix
...
     void WebSocketTransport::poll()
     {
+        if (_readyState == OPEN)
+        {
+            // if (1) ping tiemout is enabled and (2) duration exceeded the maximum delay, then close the connection
+            if (pingTimeoutExceeded())
+            {
+                close(1011, "Ping timeout");
+            }
+            // If (1) ping is enabled and (2) send for a duration exceeding our ping interval, send a ping to the server.
```
bsergean: That comment is not accurate anymore. What about making that:

```
// If (1) ping is enabled and no ping has been sent for a duration 
// exceeding our ping interval, send a ping to the server.
```

ixwebsocket/IXWebSocketTransport.cpp
bsergean: The new code tries to send pings more frequently, which is a behavior change, but is necessary if we want to make sure that ping are properly acknowledged.

Is it not possible to put that conditional block in the callback, under the 'pollResult == PollResultType::Timeout' case ? That would make the code more coherent. Would the test still pass with this ?

Or maybe we'd execute that block inconditionally (and forget that PollResultType::Timeout). But I'd like to see if we can add this code inside the callback.

test/IXWebSocketPingTest.cpp
```diff
@@ -188,36 +192,161 @@ TEST_CASE("Websocket_heartbeat", "[heartbeat]")
...
+    SECTION("Make sure that ping messages are sent when no other data are sent.")
+    {
+        ix::setupWebSocketTrafficTrackerCallback();
+
+        int port = getFreePort();
+        ix::WebSocketServer server(port);
+        std::atomic<int> serverReceivedPingMessages(0);
+        REQUIRE(startServer(server, serverReceivedPingMessages));
+
+        std::string session = ix::generateSessionId();
```
bsergean: Sometimes I create a temporary variable here to self comment things, instead of using 'false' which we don't know what that false is for.

ixwebsocket/IXWebSocketTransport.cpp
Kumamon38: Yes

ixwebsocket/IXWebSocketTransport.cpp
Kumamon38: Maybe, I’ll have to look

ixwebsocket/IXWebSocketTransport.h
Kumamon38: 😅

test/IXWebSocketPingTest.cpp
Kumamon38: There are few test case where I put true or false with a comment explaining, but we could create a var locally too

ixwebsocket/IXWebSocketTransport.cpp
Kumamon38: We need to do this in all cases so we can’t put that inside AND in the Timeout condition, because if we have always data to process the Timeout condition will never be true and there won’t be a Ping sent nor a ping timeout check. We could maybe put that code inside the poll, but before and without checking the Timeout. It should work the same, and we also have to be sure that calling close() method inside poll is not a problem. 
The test should still pass

bsergean: Thanks for the explanation. 

1. You convinced me that the timeout cannot be used, for the case when the connection is very busy and the timeout never happen. Let's forget about using it.
2. We call close from within the callback already so that should be alright to call close within the callback. -> Can you try to move that block inside the callback ? Adding a small comment, which is a hash of what you said in this review would be good.

Kumamon38: No problem, I think you answered from my original comment received in the email, I updated it since

bsergean: I think I still prefer to have this code run inside the callback. If I end up being wrong we can just make a micro-change and move it out of there back to where you have it now.

This function will be called over and over, so 'before' and 'after' state will become irrelevant fast.

Kumamon38: Ok then I will move to in the poll tomorrow

Kumamon38:
This PR is already ready to be merged with PR #34 to use the constants, if you merge #34 before, I'll push to adapt this PR.

Kumamon38:
hi,
I think about what we have done with the new timeout and I think there is still some kind of issue. The ping should be sent at a given frequency, but if we look at how it is done now, the ping could be delayed a bit.
As for example of a 1 second ping, imagine a ping sent, then you process inputs or outputs for 900ms, then nothing occurs so you will wait for the pingIntervalOrtimeoutGCsecs (1 second) whereas the next ping should be sent in just 100ms. This is clearly a problem imo.

To solve this, we have 2 possibilities:

In the poll, have a some kind of currenttime and nextdelaytime and compute the lasting delay to wait and give it in as poll timeout. In my previous example the lasting delay would be 100ms. With this solution the poll timeout parameter should be changed to millisecond and not second. I would also suggest to send the ping before the poll, that way the poll will return directly as there will be data to process.

The other solution is the thread I had the first time, and even if adding a thread is not great to you, it would clearly separate the check/send ping and the poll, and there won’t be delays as the thread is used just for this behavior.

Kumamon38:
Maybe we should be more tolerant about the delay, i.e 5ms the expected time before would be ok too

bsergean:
It would be best to add a unittest that demonstrate the problem, or to tweak one of the existing ones.

I prefer that we don't add a new thread. Having the timeout that we pass to socket->poll() adjustable based on the duration of the callback execution is a good idea. Yeah it's good to add a little error buffer, 5ms or maybe a percent such as 1% of the ping interval.

Kumamon38:
I think a new unit test would be simple to add; so we will change the timeout param from second to millisecond too

Kumamon38:
I think we also should change the closeWiresize to bool remote in the closeInfo as well as everywhere we use this

bsergean:
That closeWireSize is an implementation detail, used to track how much data flows through the socket. So we should try to hide it as much as we can, but still make sure that this data is tracked.

Users won't care about it, but maybe it can be made a private member of the closeInfo ? However that close info is a struct so nothing is hidden there.

Kumamon38:
I thought it was in the struct but not, In fact I would like be able to know in the close callback if this is local or remote closure, the close wire size != 0 would means remote currently

Kumamon38:
when we close the connection we put 0 as close wire size but maybe we should not? The Boolean param remote would make sense

bsergean:
Is the callback executed right now when the close is executed from the client (local as opposed to remote) ?

From a user pov I don't think that we want that ...

Kumamon38:
The callback is called no matter what triggered it, local or remote, as it’s triggered on the setReadState(Closed)

bsergean:
Alright. If we invoke regardless we'll need a way to distinguish remote or local closing, with a new boolean in the closeInfo as you suggest.

Kumamon38:
Great, I’ll make all the changes on Tuesday

Issue #33, opened 2019-04-17
Close description can be un-initialized at time


bsergean:
We can see this easily by running continuously `make ws_test`. In that run below it it set to Normal closure / but at other times it is set to '%'.

@Dimon4eg + @Kumamon38 / This is likely a regression of 935e6791a35a48d1f4d41b767696c2101ac1f476

I think this is why I was only setting the close reason only once in a weird way, like the first time it gets sets in one of the callbacks.

```
Received 61 bytes
ws_transfer: 0 bytes left to be sent
ws_send: received message (61 bytes)
Done !
ws_send: connection closed: code 1000 reason Normal closure

Closed connection code 1000 reason %
WebSocketServer::handleConnection() done
Received file does exists, exiting loop
1897227451 20480000 /tmp/ws_test/send/20M_file
1897227451 20480000 /tmp/ws_test/receive/20M_file
Done !
Closed connection code ws_receive: connection closed: code 1000 reason Normal closure
0
 reason 
WebSocketServer::handleConnection() done
```

another run

```
ws_send: received message (61 bytes)
ws_send: connection closed: code 1000 reason Normal closure

Closed connection code 1000 reason ??
WebSocketServer::handleConnection() done
Received file does exists, exiting loop
4293923412 20480000 /tmp/ws_test/send/20M_file
4293923412 20480000 /tmp/ws_test/receive/20M_file
Done !
ws_receive: connection closed: code 1000 reason Normal closure
Closed connection
 code 1000 reason ac
```

bsergean:
I'll take a look when I have time but if you have an idea for a fix please go ahead :)

bsergean:
ps: I should say that I am not positive that it is a regression, but it is still a bug.

Kumamon38:
I had a look and created a PR to fix this, but in fact this was nothing to do with the #28 
- First, when we receive the close code / reason remotely, the reason had 2 characters more than it needed (that why we could see "ac" or "%" or whatever randomnly read from memory)
- Second, when we used the close() method, we sent the code but not the reason, so we can't expect to read it on the other side
- Not directly linked, but we can call the `_onCloseCallback(_closeCode, _closeReason, _closeWireSize);` with not really initialized value (0, "", 0) in the case we call `setReadyState(CLOSED)` outside the `close()` method

All of this should be fixed in PR #34

Kumamon38:
So is the issue fixed as you wanted with last PR?

bsergean:
[...]

PR #34, opened 2019-04-18
fix close code/reason issue

Kumamon38:
Fix issue #33

- When `close(code, reason, wireSize)` is called, we use the close code, the close reason and send the close frame using these value, and after get back to default value.
- When the connection is closed without CLOSE frame, we use `1006` / `Abnormal closure`
- In all other cases (`setReadyState(CLOSED)`) and by default, code `1011` /  `"Internal error"` are sent when the connection is closed


bsergean:
Hey I have merged the other PR, but I wasn't too sure on which one to merge first.

Do you know how to resolve the conflicts for that one ?

bsergean:
Trying to resolve the merge myself with github UI ...

Kumamon38:
you can look at my branch real-ping-origin which is the merge of both PR

bsergean:
Thanks. I was able to do the merge myself eventually.

Kumamon38:
Great 👍 finally we removed a bug with the reason misread

PR #35, opened 2019-04-19
Link zlib statically for windows

Dimon4eg:


bsergean:
Make sense. I can't remember why I didn't take that route but it's better that way.

PR #36, opened 2019-04-19
Ping timeout use constant

Kumamon38:
- Use constants for ping timeout instead `(1011, "Ping Timeout")`
- Change type of close codes from `int` to `uint16_t`


bsergean:
Thanks !

PR #43, opened 2019-04-25
Speedup build for Windows

Dimon4eg:
On my PC with 4 cores iwx builds for Windows:
* Without `/MP` - 19 sec
* With `/MP` - 12 sec


bsergean:
Looking good, thanks !

PR #45, opened 2019-04-25
Fixes for windows

Dimon4eg:
Before fixes:
```
test cases: 3 | 3 failed
assertions: 7 | 4 passed | 3 failed
```
After fixes:
```
test cases: 3 | 2 passed | 1 failed
assertions: 9 | 8 passed | 1 failed
```


ixwebsocket/IXDNSLookup.cpp
```diff
@@ -27,7 +27,7 @@ namespace ix
         _done(false),
         _id(_nextId++)
     {
-
```
bsergean: I don't know what is the best place to put that init. I'd rather not do it here, but instead at the top of the main.
We would have to call it in the Socket() constructor too, for the people who just want to use a WebSocketServer that doesn't need to do an DNS.

bsergean: Just found this on the internet:

```
Yes, WSAStartup() can be called multiple times in a single process, as long as the requested version is supported by the WinSock DLL. Calls to WSAStartup() and WSACleanup() must be balanced.
```

If we need to balance the number of calls to WSACleanup, then we need a static variable 'initialized' or equivalent, and make sure we only ever call that function once. If we are doing this, it would be good to call call_once().

But this makes me think that it would be best to just *require* that this function is only ever called once at init time, as opposed to have us lazily call it.

bsergean:
So I merged this PR but after doing so I did commit 0a71576 / which calls that Winsock business explicitely. We can discuss in a separate PR if we want to do that lazily, but at least most of the rest of the code is merged.

PR #48, opened 2019-05-01
User/bsergeant/close with params plus close unittest

bsergean:
Trying to take some bits of PR#42 which we're having trouble to merge. Namely the unittest + passing in codes when closing.

I am able to get the same TSAN error on ubuntu xenial / I wonder if I will also get the other infinite loops.


Kumamon38:
It seems like it ends, with the same tsan errors

bsergean:
Ok I'm going to try to fix that TSAN error, which is only happening when gcc + the gcc standard library is used.

The same unittest is passing which is good.

Kumamon38:
Great, let me know if you found the issue before you sleep as I was planning to have a look tomorrow

Kumamon38:
I was also wondering if we should add the close and reason param to method stop() as well

bsergean:
Yes we should. I'll try to do that.

Kumamon38:
If so then in the CloseTest.cpp we could change the client stop from close(code, reason) and stop() to stop(code, reason)

Kumamon38:
I've been looking at this, but I cannot find yet an idea of the cause, that's quite hard :/

Kumamon38:
I think there is some trouble while accessing the `_socket` member of the WebSocketTransport.cpp, it occurs when the server close the client one by one when stopping; then it can occurs in sendOnSocket, or wakeUpFromPoll if we don't send the close frame or _socket->close() if we avoid the wakeUp

bsergean:
Yes it's strange I don't really get it.
What I don't get is that we are still using the same socket which wasn't even opened yet (), but we are still trying to close it.

I don't think that we need to had more locks here, or that we have to fix the problem inside WebSocketTransport, but instead change the structure of the server.

One last possibility is that TSAN is broken to some extent on ubuntu xenial, but I don't think so.

Another thing I noticed with my naive branch is that sometimes the macOS unittest fails locally, so it's a hint that the code is wrong (probably because I don't have some of your branch changes).

One last thing, could we still try to merge your other PR with more correct ping/pong time calculation and forget about this for now, or do we need absolutely need to merge that one first ?

Kumamon38:
Yeah I also saw that it failed locally on my mac, I possibly already fixed that in my other PR.
I think changing the server could be nice, because adding lock here would be a pain to understand.

I'll try to see if we can tomorrow but I'm not sure as I use the close with reasons if I remember well

Kumamon38:
In fact we didn’t change the behavior here so it means the master has also the problem, if we can call it like that. Maybe we could merge no?

bsergean:
I agree that there's something broken in master.

However we cannot break CI for sure, this has to stay green otherwise it can become chaos. And since this change is about close, it would be weird to merge it without a unittest around close behavior.

I think that the right course of action is first to fix that server bug. I hope I'll have some time to look at this.

Sorry it is taking so long to merge your work but those are the steps I'd like us to follow.

Kumamon38:
Yeah, sure, in fact this is only the last test case (server close) that fails, not the others.
I don’t know if we should fix before or remove just that test

Kumamon38:
Is this branch up to date with master?

bsergean:
I've merged most of the things here in master (passing code and reason) but most unittest are still commented.

PR #50, opened 2019-05-06
Fix for windows

Dimon4eg:
Now all tests passed on Windows 7 x64 & x86

### DIFF
--- a/ixwebsocket/IXSocket.cpp
+++ b/ixwebsocket/IXSocket.cpp
@@ -196,6 +196,25 @@ namespace ix
 #endif
     }
 
+    bool Socket::isWaitNeeded()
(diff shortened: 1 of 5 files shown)


ixwebsocket/IXSocket.cpp
```diff
@@ -196,6 +196,25 @@ namespace ix
...
 
+    bool Socket::isWaitNeeded()
+    {
+        int err = getErrno();
+
+        if (err == EWOULDBLOCK || err == EAGAIN)
+        {
+            return true;
+        }
+
```
bsergean: Do we need 'both tests' on Windows ? Is that to be on the safe side ?

PR #51, opened 2019-05-06
Fix windows

Dimon4eg:
More fixes for Windows.

These tests now pass:
*  IXWebSocketServerTest.cpp
*  IXWebSocketPingTest.cpp
*  IXWebSocketTestConnectionDisconnection.cpp

### DIFF
--- a/ixwebsocket/IXSocketConnect.cpp
+++ b/ixwebsocket/IXSocketConnect.cpp
@@ -7,6 +7,7 @@
 #include "IXSocketConnect.h"
 #include "IXDNSLookup.h"
 #include "IXNetSystem.h"
+#include "IXSocket.h"
 
 #include <string.h>
 #include <fcntl.h>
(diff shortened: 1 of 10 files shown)


ixwebsocket/IXSocketConnect.cpp
```diff
@@ -56,27 +45,29 @@ namespace ix
...
+                Socket::closeSocket(fd);
                 errMsg = "Cancelled";
                 return -1;
             }
 
-            // Use select to check the status of the new connection
-            struct timeval timeout;
-            timeout.tv_sec = 0;
-            timeout.tv_usec = 10 * 1000; // 10ms timeout
             fd_set wfds;
```
bsergean: The timeout needs to be re-initialized everytime / we need a comment for that. This is a problem on Linux where the select implementation 'edits the timeout'

```


       On Linux, select() modifies timeout to reflect the amount of time not
       slept; most other implementations do not do this.  (POSIX.1 permits
       either behavior.)  This causes problems both when Linux code which
       reads timeout is ported to other operating systems, and when code is
       ported to Linux that reuses a struct timeval for multiple select()s
       in a loop without reinitializing it.  Consider timeout to be unde‐
       fined after select() returns.
```

PR #52, opened 2019-05-06
added tests for IXUrlParser

Dimon4eg:
On test case fails on Windows:
```
IXWebSocket\test\IXUrlParserTest.cpp(95): FAILED:
due to unexpected exception with message:
  regex_error(error_complexity): The complexity of an attempted match against a
  regular expression exceeded a pre-set level.
```

It fails on my real url:
`ws://127.0.0.1:7350/ws?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1NTcxNzAwNzIsInVpZCI6ImMwZmZjOGE1LTk4OTktNDAwYi1hNGU5LTJjNWM3NjFmNWQxZiIsInVzbiI6InN2YmhOdlNJSmEifQ.5L8BUbpTA4XAHlSrdwhIVlrlIpRtjExepim7Yh5eEO4&status=true&format=protobuf`

Good joke about regex:
Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.
:smile:

### DIFF
--- a/test/IXUrlParserTest.cpp
+++ b/test/IXUrlParserTest.cpp
@@ -1,6 +1,6 @@
 /*
  *  IXSocketTest.cpp
- *  Author: Benjamin Sergeant
+ *  Author: Korchynskyi Dmytro
  *  Copyright (c) 2019 Machine Zone. All rights reserved.
  */
 
--- a/test/IXUrlParserTest.cpp
+++ b/test/IXUrlParserTest.cpp
@@ -1,0 +1,113 @@
+/*
+ *  IXSocketTest.cpp
+ *  Author: Benjamin Sergeant
+ *  Copyright (c) 2019 Machine Zone. All rights reserved.
+ */
+
+#include <iostream>
+#include <ixwebsocket/IXUrlParser.h>
+
+#include "IXTest.h"
+#include "catch.hpp"
+#include <string.h>
+
+using namespace ix;
+
+namespace ix
+{
+
+TEST_CASE("urlParser", "[urlParser]")
+{
+    SECTION("http://google.com")
+    {
+        std::string url = "http://google.com";
+        std::string protocol, host, path, query;
+        int port;
+        bool websocket = false;
+        bool res;
+
+        res = UrlParser::parse(url, protocol, host, path, query, port, websocket);
+
+        REQUIRE(res);
+        REQUIRE(protocol == "http");
+        REQUIRE(host == "google.com");
+        REQUIRE(path == "/");
+        REQUIRE(query == "");
+        REQUIRE(port == 80); // default port for http
+    }
+
+    SECTION("https://google.com")
+    {
+        std::string url = "https://google.com";
+        std::string protocol, host, path, query;
+        int port;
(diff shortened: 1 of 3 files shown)


test/IXUrlParserTest.cpp
```diff
@@ -0,0 +1,113 @@
+/*
+ *  IXSocketTest.cpp
+ *  Author: Benjamin Sergeant
```
bsergean: Don't you want to add your name in there for posterity ?

Dimon4eg: Oki

Dimon4eg:
On my PC with core i7 CPU it parses several seconds and then fails.
All rest cases work immediately.

Dimon4eg:
It works on Mac and Linux :roll_eyes: 
@bsergean  What do you suggest to fix this issue on Windows?

bsergean:
Ideally we would tweak the regex a little bit and find one that doesn't give trouble to the Microsoft implementation. Now that we have a unittest it should help.

We can also kill the regexp, and write some manual parsing. If we do that we can get rid of the static member variable etc...

bsergean:
http://www.zedwood.com/article/cpp-boost-url-regex

I think this is what I used a while ago to implement this. They have a 'no-regexp' implementation which is small and a bit cryptic. We could try to swap the regexp with this:

```

void parse_url(const string& raw_url) //no boost
{
    string path,domain,x,protocol,port,query;
    int offset = 0;
    size_t pos1,pos2,pos3,pos4;
    x = _trim(raw_url);
    offset = offset==0 && x.compare(0, 8, "https://")==0 ? 8 : offset;
    offset = offset==0 && x.compare(0, 7, "http://" )==0 ? 7 : offset;
    pos1 = x.find_first_of('/', offset+1 );
    path = pos1==string::npos ? "" : x.substr(pos1);
    domain = string( x.begin()+offset, pos1 != string::npos ? x.begin()+pos1 : x.end() );
    path = (pos2 = path.find("#"))!=string::npos ? path.substr(0,pos2) : path;
    port = (pos3 = domain.find(":"))!=string::npos ? domain.substr(pos3+1) : "";
    domain = domain.substr(0, pos3!=string::npos ? pos3 : domain.length());
    protocol = offset > 0 ? x.substr(0,offset-3) : "";
    query = (pos4 = path.find("?"))!=string::npos ? path.substr(pos4+1) : "";
    path = pos4!=string::npos ? path.substr(0,pos4) : path;
    cout << "[" << raw_url << "]" << endl;
    cout << "protocol: " << protocol << endl;
    cout << "domain: " << domain << endl;
    cout << "port: " << port << endl;
    cout << "path: " << path << endl;
    cout << "query: " << query << endl;
}
```

bsergean:
It's from that page -> http://www.zedwood.com/article/cpp-boost-url-regex

Dimon4eg:
I'd rather use more robust and tested lib: https://github.com/corporateshark/LUrlParser

PR #54, opened 2019-05-06
update README.md

Dimon4eg:


bsergean:
A good next step would be to figure out how we can setup CI on travis.

PR #49, opened 2019-05-03
fix ping, fix send frame close

Kumamon38:
- Fix ping timeout
- Fix an issue about callback/code read from remote.
When we close the connection using close(code, reason...), a CLOSE frame is sent to the remote, and then right after we really close the socket connection. On the remote side, the CLOSE frame data are read, then the connection is immediately closed so the close callback is called with 1006 / Abnormal close instead of the values of the CLOSE frame, because these data are not processed in the dispatch yet. The fix is to close the connection, set that we need to read the buffer, and if nothing triggered a CLOSED state with the data read, then call the callback with 1006 / Abnormal close

PR #58, opened 2019-05-08
fix close reason when closing by self

Dimon4eg:
Some times when closing by self I receive close notification with abnormal close reason.
This PR fixes this issue.


bsergean:
@Kumamon38 is working on something similar in a different PR. Does that fix look correct ?

Dimon4eg:
Closing because @Kumamon38 has fixed this in PR #49

PR #59, opened 2019-05-08
Added WebSocketMessageQueue class

Dimon4eg:
Added helper class with `poll` functionality.

### DIFF
--- a/ixwebsocket/IXWebSocket.h
+++ b/ixwebsocket/IXWebSocket.h
@@ -113,6 +113,11 @@ namespace ix
         WebSocketSendInfo ping(const std::string& text);
         void close();
 
+        /**
+         * Set callback to receive websocket messages.
+         *
+         * Be aware: your callback will be executed from websocket's internal thread!
+         */
         void setOnMessageCallback(const OnMessageCallback& callback);
         static void setTrafficTrackerCallback(const OnTrafficTrackerCallback& callback);
         static void resetTrafficTrackerCallback();
--- a/ixwebsocket/IXWebSocket.h
+++ b/ixwebsocket/IXWebSocket.h
@@ -113,12 +113,10 @@ namespace ix
         WebSocketSendInfo ping(const std::string& text);
         void close();
 
-        /**
-         * Set callback to receive websocket messages.
-         *
-         * Be aware: your callback will be executed from websocket's internal thread!
-         */
+        // Set callback to receive websocket messages.
+        // Be aware: your callback will be executed from websocket's internal thread!
         void setOnMessageCallback(const OnMessageCallback& callback);
+
         static void setTrafficTrackerCallback(const OnTrafficTrackerCallback& callback);
         static void resetTrafficTrackerCallback();
 
--- a/ixwebsocket/IXWebSocketMessageQueue.h
+++ b/ixwebsocket/IXWebSocketMessageQueue.h
@@ -1,7 +1,7 @@
 /*
- *  IXWebSocketPoll.h
+ *  IXWebSocketMessageQueue.h
  *  Author: Korchynskyi Dmytro
- *  Copyright (c) 2017-2018 Machine Zone, Inc. All rights reserved.
+ *  Copyright (c) 2017-2019 Machine Zone, Inc. All rights reserved.
  */
 
 #pragma once
@@ -13,9 +13,9 @@
 
 namespace ix
 {
-    /**
-     * A helper class to dispatch websocket message callbacks in your thread.
-     */
+    //
+    // A helper class to dispatch websocket message callbacks in your thread.
+    //
     class WebSocketMessageQueue
     {
     public:
--- a/ixwebsocket/IXWebSocketMessageQueue.h
+++ b/ixwebsocket/IXWebSocketMessageQueue.h
@@ -39,14 +39,14 @@ namespace ix
             WebSocketCloseInfo closeInfo;
         };
 
-        using MessageDataPtr = std::shared_ptr<Message>;
+        using MessagePtr = std::shared_ptr<Message>;
 
-        MessageDataPtr popMessage();
+        MessagePtr popMessage();
 
     private:
         WebSocket* _websocket = nullptr;
         OnMessageCallback _onMessageUserCallback;
         std::mutex _messagesMutex;
-        std::list<MessageDataPtr> _messages;
+        std::list<MessagePtr> _messages;
     };
 }
--- a/ixwebsocket/IXWebSocketMessageQueue.h
+++ b/ixwebsocket/IXWebSocketMessageQueue.h
@@ -16,11 +16,11 @@ namespace ix
     /**
      * A helper class to dispatch websocket message callbacks in your thread.
      */
-    class WebSocketPoll
+    class WebSocketMessageQueue
     {
     public:
-        WebSocketPoll(WebSocket* websocket = nullptr);
-        ~WebSocketPoll();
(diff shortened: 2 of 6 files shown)


ixwebsocket/IXWebSocket.h
```diff
@@ -113,6 +113,11 @@ namespace ix
         WebSocketSendInfo ping(const std::string& text);
         void close();
 
+        /**
```
bsergean: I'm using // more tidy comment for now, not the doxygen style. I'd like to stick to the current function style.

ixwebsocket/IXWebSocketPoll.cpp
```diff
@@ -0,0 +1,109 @@
...
+ *  IXWebSocketPoll.cpp
+ *  Author: Korchynskyi Dmytro
+ *  Copyright (c) 2017-2018 Machine Zone, Inc. All rights reserved.
+ */
+
+#include "IXWebSocketPoll.h"
+
+namespace ix
+{
+
```
bsergean: Could we use a shared_ptr instead of a raw pointer?
Those make me nervous.

ixwebsocket/IXWebSocketPoll.cpp
```diff
@@ -0,0 +1,109 @@
...
+        if (!_messages.empty())
+        {
+            // not handled all messages
+        }
+
+        bindWebsocket(nullptr);
+    }
+
+    void WebSocketPoll::bindWebsocket(WebSocket * websocket)
+    {
```
bsergean: I don't like code that is too indented if I can avoid it. Here I would prefer the early out style:

```
if (_websocket == websocket) return;
```

ixwebsocket/IXWebSocketPoll.cpp
```diff
@@ -0,0 +1,109 @@
...
+        {
+            // unbind old
+            if (_websocket)
+            {
+                _websocket->setOnMessageCallback(nullptr);
+            }
+
+            _websocket = websocket;
+
+            // bind new
```
bsergean: With a shared_ptr we probably don't need this null check right ?

ixwebsocket/IXWebSocketPoll.cpp
```diff
@@ -0,0 +1,109 @@
...
+                    MessageDataPtr message(new Message());
+
+                    message->type      = type;
+                    message->str       = str;
+                    message->wireSize  = wireSize;
+                    message->errorInfo = errorInfo;
+                    message->openInfo  = openInfo;
+                    message->closeInfo = closeInfo;
+
+                    _messagesMutex.lock();
```
bsergean: I'd like to avoid explicit mutex lock and unlock, and instead use lock_guard which works better with exceptions, and take less code (no need for the explicit unlock)

ixwebsocket/IXWebSocketPoll.cpp
```diff
@@ -0,0 +1,109 @@
...
+    void WebSocketPoll::setOnMessageCallback(const OnMessageCallback& callback)
+    {
+        _onMessageUserCallback = callback;
+    }
+    
+    WebSocketPoll::MessageDataPtr WebSocketPoll::popMessage()
+    {
+        MessageDataPtr message;
+
+        _messagesMutex.lock();
```
bsergean: same mutex comment

ixwebsocket/IXWebSocketPoll.h
```diff
@@ -0,0 +1,52 @@
+/*
+ *  IXWebSocketPoll.h
+ *  Author: Korchynskyi Dmytro
+ *  Copyright (c) 2017-2018 Machine Zone, Inc. All rights reserved.
```
bsergean: 2019

ixwebsocket/IXWebSocketPoll.cpp
```diff
@@ -0,0 +1,109 @@
+/*
+ *  IXWebSocketPoll.cpp
+ *  Author: Korchynskyi Dmytro
+ *  Copyright (c) 2017-2018 Machine Zone, Inc. All rights reserved.
```
bsergean: 2019

Dimon4eg:
WebSocketMessageQueue sounds good more

ixwebsocket/IXWebSocketPoll.cpp
Dimon4eg: No, because user might not use shared_ptr

ixwebsocket/IXWebSocketPoll.cpp
Dimon4eg: Still need because shared_ptr also can be null

ixwebsocket/IXWebSocketPoll.cpp
Dimon4eg: Ok

ixwebsocket/IXWebSocketPoll.cpp
Dimon4eg: Ok

ixwebsocket/IXWebSocketPoll.cpp
Dimon4eg: Ok

ixwebsocket/IXWebSocketMessageQueue.h
```diff
@@ -0,0 +1,52 @@
...
+        ~WebSocketMessageQueue();
+
+        void bindWebsocket(WebSocket* websocket);
+
+        void setOnMessageCallback(const OnMessageCallback& callback);
+
+        void poll(int count = 512);
+
+    protected:
+        struct Message
```
Dimon4eg: @bsergean what about using the Message struct in IXW and pass it as shared_ptr?
This will eliminate redundant copy of data which might be long.

bsergean: Yes I was thinking about that. This would simplify all APIs to only take a 'BigMessage'. We'd have to put that struct or class into its own header file.

ixwebsocket/IXWebSocketPoll.cpp
bsergean: Can we use a const reference to an IXWebSocket ?

ixwebsocket/IXWebSocketMessageQueue.h
bsergean: BTW we should do that in a different PR, because this will be a very messy change.

ixwebsocket/IXWebSocketPoll.cpp
Dimon4eg: No, we set callback on it

ixwebsocket/IXWebSocketMessageQueue.h
Dimon4eg: Totally agree

bsergean: Also we should think about whether this will create more allocs when copying those small Messages (mostly for the str).

Dimon4eg: string in any case allocates memory

Dimon4eg: for short string might not, this depends on string implementation

bsergean:
This is looking good. Before merging it would be good to have a simple unittest, that start a server and a client, where we send N messages and make sure we can retrieve them by calling poll.

Dimon4eg:
Okay

PR #61, opened 2019-05-09
Fail test GCC TSAN

Kumamon38:
On Linux with GCC there are TSAN errors, the data race seems to happen the first time the thread has enough time to successfully connect (after handshake). After the successful, it's ok, and before each attempt stopped before success, it's ok too. The index where the change occurs depends on the running computer, ie on Travis it's Run 3, on mine it's Run 8
We could reduce the test loop to this too:
```
SECTION("Try to connect to invalid servers.")
    {
        IXWebSocketTestConnectionDisconnection chatA;
        // not enough time to connect
        chatA.start(WEBSOCKET_DOT_ORG_URL);
        ix::msleep(7*50);
        chatA.stop();
        // enough time to connect
        chatA.start(WEBSOCKET_DOT_ORG_URL);
        ix::msleep(8*50);
        chatA.stop();
    }
```
the error will appear

### DIFF
--- a/test/IXWebSocketTestConnectionDisconnection.cpp
+++ b/test/IXWebSocketTestConnectionDisconnection.cpp
@@ -126,7 +126,7 @@ TEST_CASE("websocket_connections", "[websocket]")
         }
     }
 
-    SECTION("Try to connect and disconnect with different timing, from not enough time to successfull connect")
+    /*SECTION("Try to connect and disconnect with different timing, from not enough time to successfull connect")
     {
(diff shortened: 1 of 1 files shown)


bsergean:
It's great that you have a very simple case that display the error. I like this PR.

But I cannot merge something that will break CI / can you comment it out the new unittest ?

Kumamon38:
Sure

Kumamon38:
Ready to be merged

PR #63, opened 2019-05-10
Improve calculateRetryWaitMilliseconds

Dimon4eg:


ixwebsocket/IXWebSocket.cpp
```diff
@@ -13,16 +13,21 @@
...
-        // and we'll be constantly retrying to connect.
-        uint64_t wait_time = ((uint64_t) std::pow(2, retry_count) * 100L);
+        uint64_t wait_time;
 
-        // cap the wait time to 10s, or to retry_count == 10 for which wait_time > 10s
-        uint64_t tenSeconds = 10 * 1000;
-        return (wait_time > tenSeconds || retry_count > 10) ? tenSeconds : wait_time;
+        if (retry_count <= 6)
+        {
+            // max wait_time is 6400
```
bsergean: max wait_time is 6400ms

ixwebsocket/IXWebSocket.cpp
```diff
@@ -13,16 +13,21 @@
 
 namespace
 {
-    uint64_t calculateRetryWaitMilliseconds(uint64_t retry_count)
+    uint64_t calculateRetryWaitMilliseconds(uint32_t retry_count)
     {
-        // This will overflow quite fast for large value of retry_count
```
bsergean: Can you keep that comment about overflow for the next programmer who come here ?

Dimon4eg: No overflow now possible as I'm checking before: `if (retry_count <= 6)`

ixwebsocket/IXWebSocket.cpp
Dimon4eg: That comments no more correct or am I wrong?

bsergean: Add 'ms'.

ixwebsocket/IXWebSocket.cpp
bsergean: Sure, but we should explain why we have this kind of odd magic value (6). Normally we wouldn't have to do that and would just use the pow(2) thing.

PR #64, opened 2019-05-11
Fix crash during closing on Windows

Dimon4eg:
It crashes when `createSocket` fails to create socket on Windows because no SSL support.

Would be good to simulate error when creating socket.
@bsergean how to do that?

### DIFF
--- a/ixwebsocket/IXWebSocket.cpp
+++ b/ixwebsocket/IXWebSocket.cpp
@@ -151,15 +151,12 @@ namespace ix
 
         close();
 
-        if (!_thread.joinable())
+        if (_thread.joinable())
         {
-            _automaticReconnection = automaticReconnection;
-            return;
+            _stop = true;
+            _thread.join();
+            _stop = false;
         }
-
-        _stop = true;
-        _thread.join();
-        _stop = false;
 
         _automaticReconnection = automaticReconnection;
     }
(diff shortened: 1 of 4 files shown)


bsergean:
The gcc + tsan ci is failing

https://travis-ci.com/machinezone/IXWebSocket/jobs/199426682

It should be possible to reproduce the problem locally if you have docker installed.

bsergean:
I think that the background thread is still active not fully joined, but we are doing some work on the main/control thread which is closing.

It's a race condition between the 2 threads I believe.

bsergean:
I checked out your branch and like your new code structure better than the previous one.
I have small comments but I'll do them myself without changing much.

If we want to override the 'socket creation', maybe we could have a factory to do that. But I don't know if that's really worth it ... we should probably add SSL support on Windows instead :)

Dimon4eg:
SSL yes needed but in case would be good to have such test

bsergean:
On another note, the unittest went from 10s to 22s to execute. Not sure what happened maybe it's more correct now, but it's annoying for developer productivity to wait a long time for unittest to finish.

Execution time: 22.27s

bsergean:
Actually maybe it's because I'm on my home DSL which is bad.

Dimon4eg:
On travis max execution time is 2:20 so it's your DSL is bad.
Why do you use it? :thinking: 
I thought it's dead nowadays.

bsergean:
We have not updated our internet provider for a long time so we still have an old router from the ISP, but since it 'works' right now I prefer not to change it. Maybe one day.

PR #67, opened 2019-05-11
use C++11 enums

Dimon4eg:
This changes public API!


bsergean:
Thanks for fixing this, C++11 enums are better than old school ones.

Can you please update the README.md as well, I think it's the only place that needs to be updated left ?

Dimon4eg:
Done. Next version I think should be 2.x

bsergean:
Sure, to reflect the big api change I guess. I added a v2.0.0 tag.

PR #68, opened 2019-05-11
Message queue

Dimon4eg:


Dimon4eg:
@bsergean tsan detects some problem during executing my new test

Kumamon38:
But the CI reports à segmentation fault for Linux GCC 🧐

Dimon4eg:
@Kumamon38 thanks, forgot about that, fixed now

ixwebsocket/IXWebSocketMessageQueue.cpp
```diff
@@ -33,12 +33,12 @@ namespace ix
         {
             // set dummy callback just to avoid crash
             _websocket->setOnMessageCallback([](
-                WebSocketMessageType type,
-                const std::string & str,
-                size_t wireSize,
-                const WebSocketErrorInfo & errorInfo,
-                const WebSocketOpenInfo & openInfo,
-                const WebSocketCloseInfo & closeInfo)
+                WebSocketMessageType,
```
bsergean: Nitpick style comment / there should be no space between type and &.

Dimon4eg: Do you use some tool to check style?

Dimon4eg:
SUMMARY: ThreadSanitizer: data race memory:3834 in ix::Socket::wakeUpFromPoll(unsigned char)

Dimon4eg:
Cool! Let's use it

bsergean:
So I merged this, and then reverted it ... because the unittest was failing.

I spent some time looking at it: it was failing because of the getClients() in the server code. I wonder if it would go away if we were not sending data back to its own socket.

```
==23785== Thread 7 :
==23785== Use of uninitialised value of size 8
==23785==    at 0x56A8C00: pthread_mutex_lock (in /usr/lib64/libpthread-2.17.so)
==23785==    by 0x409626: __gthread_mutex_lock(pthread_mutex_t*) (gthr-default.h:748)
==23785==    by 0x442ECD: std::mutex::lock() (mutex:135)
==23785==    by 0x456A01: std::lock_guard<std::mutex>::lock_guard(std::mutex&) (mutex:386)
==23785==    by 0x54201E: ix::WebSocketServer::getClients() (IXWebSocketServer.cpp:118)
```

bsergean:
I did add back the files to master, but I have disabled the unittest. It would be good to figure out what's wrong with it, as it is possibly uncovering a bug.

PR #69, opened 2019-05-11
fix for Windows

Dimon4eg:


ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -303,6 +303,8 @@ namespace ix
             }
         }
 
+        if (lastingTimeoutDelayInMs) lastingTimeoutDelayInMs = 20;
```
bsergean: That one looks like debug code (overriding to 20ms), or do you want to always timeout after 20ms on Windows ? That could solve some problems since the select interrupt is a no-op on Windows.

ixwebsocket/IXSocket.cpp
```diff
@@ -73,7 +73,7 @@ namespace ix
 
         struct timeval timeout;
         timeout.tv_sec = timeoutMs / 1000;
-        timeout.tv_usec = (timeoutMs < 1000) ? 0 : 1000 * (timeoutMs % 1000);
+        timeout.tv_usec = 1000 * (timeoutMs % 1000);
```
bsergean: That's good / I wonder how we ended up with that.

Dimon4eg: Me too :+1: )

PR #71, opened 2019-05-12
Fix run.py

Dimon4eg:
Fixed:
* Wrong error output `Unittest timed out` 
* Windows support

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 build
+*.pyc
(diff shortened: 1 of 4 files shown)


Dimon4eg:
Sometimes test fails:
```
/home/travis/build/machinezone/IXWebSocket/test/IXWebSocketPingTest.cpp:423: FAILED:
  REQUIRE( serverReceivedPingMessages == 1 )
with expansion:
  2 == 1
```

@Kumamon38 should fix it

bsergean:
Hey we are not building in parallel now it's a bit annoying at CI takes longer to run.

I tried --parallel flag to cmake but this isn't working too well. Maybe I'll bring back the old way for building Unix, and keep `cmake --build` for windows

Issue #70, opened 2019-05-12
Ping unittests are broken on the master branch


Dimon4eg:
SUMMARY: ThreadSanitizer: data race on vptr (ctor/dtor vs virtual call) IXSelectInterrupt.cpp:17 in ix::SelectInterrupt::~SelectInterrupt()

https://travis-ci.org/machinezone/IXWebSocket/jobs/531244480

/Users/travis/build/machinezone/IXWebSocket/test/IXWebSocketPingTest.cpp:473: FAILED:
  REQUIRE( serverReceivedPingMessages == 2 )
with expansion:
  3 == 2
https://travis-ci.com/machinezone/IXWebSocket/jobs/199446279

Both fail only on Mac

bsergean:
Just pushed something to make all derived classes use 'final' keyword.

Hope it helps, it is recommended here for the 'vptr' kindof error.
https://github.com/google/sanitizers/wiki/ThreadSanitizerPopularDataRaces

Dimon4eg:
It seems now it's fixed!

bsergean:
I think that the second one might still be broken, and it will fail randomly. That one.

/Users/travis/build/machinezone/IXWebSocket/test/IXWebSocketPingTest.cpp:473: FAILED:
REQUIRE( serverReceivedPingMessages == 2 )
with expansion:
3 == 2
https://travis-ci.com/machinezone/IXWebSocket/jobs/199446279

Dimon4eg:
Still fails:
```
/Users/travis/build/machinezone/IXWebSocket/test/IXWebSocketPingTimeoutTest.cpp:368: FAILED:
  REQUIRE( webSocketClient.isClosed() == true )
with expansion:
  false == true
```

bsergean:
FYI I have disabled the 2 Ping related unittest, which isn't good, but better than super flaky unittest. It still fails somewhat randomly at times, but looks more stable now.

I've made a server change for the best I hope.

Kumamon38:
What do we have to do then? We are not sure about the CI or the code

bsergean:
I'm not too sure either. Here are some random thoughts:

We could have a branch / or PR where we tweak things until something gets stable.
The key with fixing a problem on the CI is being able to reproduce it locally.
Maybe there's a more stable way to write tests with timing.
[...]

PR #72, opened 2019-05-13
save timepoints after connect and not in contructor, adjusted tests

Kumamon38:
I think this should fix the random error of Issue #70

### DIFF
--- a/ixwebsocket/IXWebSocketTransport.cpp
+++ b/ixwebsocket/IXWebSocketTransport.cpp
@@ -168,7 +168,6 @@ namespace ix
                                                          timeoutSecs);
         if (result.success)
         {
-            initTimePointsAndGCDAfterConnect();
             setReadyState(ReadyState::OPEN);
         }
         return result;
@@ -197,7 +196,6 @@ namespace ix
         auto result = webSocketHandshake.serverHandshake(fd, timeoutSecs);
         if (result.success)
         {
-            initTimePointsAndGCDAfterConnect();
             setReadyState(ReadyState::OPEN);
         }
         return result;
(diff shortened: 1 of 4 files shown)


ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -234,6 +231,23 @@ namespace ix
...
+            _lastSendPingTimePoint = std::chrono::steady_clock::now();
+        } 
+        {
+            std::lock_guard<std::mutex> lock(_lastReceivePongTimePointMutex);
+            _lastReceivePongTimePoint = std::chrono::steady_clock::now();
+        }
+
+        if (_pingIntervalOrTimeoutGCDSecs > 0)
+        {
+            _nextGCDTimePoint = std::chrono::steady_clock::now() + std::chrono::seconds(_pingIntervalOrTimeoutGCDSecs);
```
Dimon4eg: The same calculation is in `configure` method. Better to move it to new method to avoid duplication of code

Kumamon38: I already removed it from the configure method

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -173,6 +168,7 @@ namespace ix
                                                          timeoutSecs);
         if (result.success)
         {
+            initTimePointsAndGCDAfterConnect();
```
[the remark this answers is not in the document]
Kumamon38: yep

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -201,6 +197,7 @@ namespace ix
         auto result = webSocketHandshake.serverHandshake(fd, timeoutSecs);
         if (result.success)
         {
+            initTimePointsAndGCDAfterConnect();
```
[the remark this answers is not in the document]
Kumamon38: yep

ixwebsocket/IXWebSocketTransport.cpp
Dimon4eg: okay

bsergean:
Ah it broke again after merging (on macOS)

https://travis-ci.com/machinezone/IXWebSocket/jobs/199740940

It is still unreliable.

PR #73, opened 2019-05-13
add more time to let client close

Kumamon38:


bsergean:
The linux ci still failed -> https://travis-ci.org/machinezone/IXWebSocket/jobs/531879796

bsergean:
I think the best way is to use condition variables that gets notified when things are ready, instead of adding ix::msleep()

Kumamon38:
Anyway it’s strange to have only 1 ping here whereas we should have 2 for a piece of time

bsergean:
Merged this but that test is disabled now. Maybe we could have a branch with those tests enabled, and try to fix things there.

PR #79, opened 2019-05-17
let poll do his job when closing

Kumamon38:
It's like before, the same bug I fixed came back :)
We have to let the poll() do his job if we are in the closing state, there won't be a poll waiting forever has a little timeout is set in Transport for state != OPEN
Btw here no callback was called at all before the fix

### DIFF
--- a/ixwebsocket/IXWebSocket.cpp
+++ b/ixwebsocket/IXWebSocket.cpp
@@ -293,8 +293,8 @@ namespace ix
                 break;
             }
 
-            // We cannot enter poll which might block forever if we are stopping
-            if (_stop) break;
+            // We can avoid to poll if we want to stop and are not closing
+            if (_stop && !isClosing()) break;
 
             // 2. Poll to see if there's any new data available
             WebSocketTransport::PollResult pollResult = _ws.poll();
--- a/ixwebsocket/IXWebSocketTransport.cpp
+++ b/ixwebsocket/IXWebSocketTransport.cpp
@@ -71,7 +71,7 @@ namespace ix
     const int WebSocketTransport::kDefaultPingIntervalSecs(-1);
     const int WebSocketTransport::kDefaultPingTimeoutSecs(-1);
     const bool WebSocketTransport::kDefaultEnablePong(true);
-    const int WebSocketTransport::kClosingMaximumWaitingDelayInMs(200);
+    const int WebSocketTransport::kClosingMaximumWaitingDelayInMs(300);
     constexpr size_t WebSocketTransport::kChunkSize;
 
     WebSocketTransport::WebSocketTransport() :
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -37,6 +37,7 @@ set (SOURCES
   IXWebSocketTestConnectionDisconnection.cpp
   IXUrlParserTest.cpp
   IXWebSocketServerTest.cpp
+  IXWebSocketCloseTest.cpp
 )
 
 # Some unittest don't work on windows yet
@@ -50,7 +51,6 @@ endif()
 # Disable tests for now that are failing or not reliable
 # IXWebSocketPingTest.cpp
 # IXWebSocketPingTimeoutTest.cpp
-# IXWebSocketCloseTest.cpp
 # IXWebSocketMessageQTest.cpp (trigger a segfault on Linux)
 
 add_executable(ixwebsocket_unittest ${SOURCES})
--- a/test/IXWebSocketCloseTest.cpp
+++ b/test/IXWebSocketCloseTest.cpp
@@ -102,6 +102,7 @@ namespace
         }
 
         _webSocket.setUrl(url);
+        _webSocket.disableAutomaticReconnection();
 
         std::stringstream ss;
         log(std::string("Connecting to url: ") + url);
@@ -118,8 +119,6 @@ namespace
                 if (messageType == ix::WebSocketMessageType::Open)
                 {
                     log("client connected");
-
-                    _webSocket.disableAutomaticReconnection();
                 }
                 else if (messageType == ix::WebSocketMessageType::Close)
                 {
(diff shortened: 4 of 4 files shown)


Kumamon38:
For the other tests, see later =)

Kumamon38:
Seems like the new Windows CI is not happy and run forever :/

Dimon4eg:
```
Running cmake --build --parallel C:\Users\travis\build\machinezone\IXWebSocket\test\build\Windows
Using 2 cores to execute the unittest
The command "python test/run.py" exited with 1.
Done. Your build exited with 1.
Traceback (most recent call last):
  File "test/run.py", line 510, in <module>
  File "test/run.py", line 506, in main
  File "test/run.py", line 409, in run
  File "test/run.py", line 278, in executeJobs
  File "C:\ProgramData\chocolatey\lib\mingw\tools\install\mingw64\opt\lib\python2.7/multiprocessing/pool.py", line 251, in map
    return self.map_async(func, iterable, chunksize).get()
  File "C:\ProgramData\chocolatey\lib\mingw\tools\install\mingw64\opt\lib\python2.7/multiprocessing/pool.py", line 558, in get
    raise self._value
WindowsError: [Error 32] The process cannot access the file because it is being used by another process: 'c:\\users\\travis\\appdata\\local\\temp\\tmpfp8rpqWebsocket_client_close_params_given.log'
```

ixwebsocket/IXWebSocket.cpp
```diff
@@ -293,8 +293,8 @@ namespace ix
                 break;
             }
 
-            // We cannot enter poll which might block forever if we are stopping
-            if (_stop) break;
+            // We can avoid to poll if we want to stop and are not closing
+            if (_stop && !isClosing()) break;
```
bsergean: The bug I fixed the other day (Wednesday) explicitely want that break clause WHEN closing, otherwise we were hanging on macOS on the CI.

Kumamon38: It seems not, the macOS Ci is green

Kumamon38: And on my Mac too :)

Kumamon38: Anyway if we don’t do this, no close callback at all are called, because we are closing and expected to deal with this situation in the poll or in the dispatch

bsergean: If no close callbacks are ever called indeed that's a serious bug. I will test that again. 
The CI has been extremely unstable, the macOS CI would randomly fail.

bsergean: I pushed that commit to run the windows CI unittest on one thread 
aa3f201cedfb4bc0697b1552eacd35afb88313b5

That will let us see the error message in the close unittest.

Kumamon38: That’s why the test fails with 0 as client close code, which means not close code set in the test so no close callback called; confirmed by no “client disconnect” test log.
That’s the problem of disabling tests and adding break conditions, before going deeper in the investigation I expected some kind of trouble like that. We don’t have to forget that close() method just send the close frame and set to closing state, but we have to wait for the close frame of the server or deals with a connection closed, so need to poll and dispatch, and if finally nothing is received the closeMaximim delay will close the connection.

Kumamon38: Will it rerun again?

bsergean:
If you rebase your branch of master and pick that change, it should re-run again and show the Windows error.

We could disable that test on windows for now.

Kumamon38:
In progress

bsergean:
The test just failed on Linux that one time.

 <>[2019-05-17 23:08:46.902] [info] client disconnected
 <>
 <>~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 <>ixwebsocket_unittest is a Catch v2.7.1 host application.
 <>Run with -? for options
 <>
 <>-------------------------------------------------------------------------------
 <>Websocket_client_close_default
 <>  Make sure that close code and reason was used and sent to server.
 <>-------------------------------------------------------------------------------
 <>/Users/travis/build/machinezone/IXWebSocket/test/IXWebSocketCloseTest.cpp:231
 <>...............................................................................
 <>
 <>/Users/travis/build/machinezone/IXWebSocket/test/IXWebSocketCloseTest.cpp:266: FAILED:
 <>  REQUIRE( webSocketClient.getCloseCode() == 1000 )
 <>with expansion:
 <>  1011 == 1000 (0x3e8)
 <>
 <>===============================================================================
 <>test cases: 1 | 1 failed
 <>assertions: 3 | 2 passed | 1 failed
 <>
 <>
 <>unittest failed
 <>Execution time: 29.64s

Kumamon38:
The client seems to be connected, but 1011 means nothing set a close code before switching to closed state, interesting

bsergean:
BTW that one is a unittest failure, which isn't as bad as a crash as we used to get.

Kumamon38:
Yeah, I’ll look at it on Monday

Kumamon38:
Windows CI fails again, if the file is lock maybe it never ends due to the poll or so merging like that

Kumamon38:
Not great to add/remove logs, but so far I've been able to find this
```
[2019-05-20 14:31:34.816] [info] client disconnected(1000,Normal closure)
[2019-05-20 14:31:34.817] [info] Server closed connection(1011,Internal error)
```
For me it means that the client didn't have time to send the close frame
We probably don't wait enough or something like that

ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -1042,22 +1042,41 @@ namespace ix
...
-        sendCloseFrame(code, reason);
+        // connection is opened, so close without sending close frame
+        if (_readyState == ReadyState::OPEN)
         {
-            std::lock_guard<std::mutex> lock(_closeDataMutex);
-            _closeCode = code;
-            _closeReason = reason;
-            _closeWireSize = closeWireSize;
-            _closeRemote = remote;
+            {
```
bsergean: that block of code (_closeCode = ...) could be put outside of the branch

PR #80, opened 2019-05-21
Fix ping

Kumamon38:
Needs merge of pr #79


bsergean:
Ok let's give the new code a try.

Hopefully the extra checks will make it more robust.

PR #105, opened 2019-08-25
Add functionality to include non standard headers in client handshake protocol

ozychhi:
Even though RFC 6455 clearly defines necessary headers to complete the handshake, in practice most of the websocket servers expect/require non-standard headers, therefore adding this functionality. Mind the fact that this includes reformatting quite a bit of code; not sure why this code was not formatted according to your `.clang-format`, I can exclude relevant commit, if you have certain preferences.


bsergean:
Thanks for the fix. Other libraries have the options to set extra headers so it sounds like a nice improvement.

1. Can you remove the 'clang-format changes' please ? It makes that PR a bit massive while it just add a few things.
2. Can you add a note in the README to document this new api.

ozychhi:
@bsergean Hey, I picked out format commit; now the changes should be minimal. Though, on the side, I would really recommend enforcing clang-formatted code (especially if you have .clang-format in your repository)

PR #114, opened 2019-09-22
WIP: support configurable certificates/keys, and root trust CAs

matt-deboer:
This adds a `SocketTLSOptions` type that can be set on clients and servers to configure the certificate used, as well as peer verification options.

### DIFF
--- a/ixwebsocket/IXSocketOpenSSL.cpp
+++ b/ixwebsocket/IXSocketOpenSSL.cpp
@@ -7,14 +7,12 @@
  */
 
 #include "IXSocketOpenSSL.h"
+
 #include "IXSocketConnect.h"
-
 #include <cassert>
-
-#include <openssl/x509v3.h>
-
-#include <fnmatch.h>
 #include <errno.h>
+#include <fnmatch.h>
+#include <openssl/x509v3.h>
 #define socketerrno errno
 
 namespace ix
@@ -22,9 +20,10 @@ namespace ix
     std::atomic<bool> SocketOpenSSL::_openSSLInitializationSuccessful(false);
     std::once_flag SocketOpenSSL::_openSSLInitFlag;
 
-    SocketOpenSSL::SocketOpenSSL(const SocketTLSOptions& tlsOptions, int fd) : Socket(fd),
-        _ssl_connection(nullptr),
-        _ssl_context(nullptr)
+    SocketOpenSSL::SocketOpenSSL(const SocketTLSOptions& tlsOptions, int fd)
+        : Socket(fd)
+        , _ssl_connection(nullptr)
+        , _ssl_context(nullptr)
     {
         std::call_once(_openSSLInitFlag, &SocketOpenSSL::openSSLInitialize, this);
     }
@@ -114,13 +113,12 @@ namespace ix
         SSL_CTX* ctx = SSL_CTX_new(_ssl_method);
         if (ctx)
         {
-            if (!_tlsOptions.isPeerVerifyDisabled()) {
+            if (!_tlsOptions.isPeerVerifyDisabled())
+            {
                 // To skip verification, pass in SSL_VERIFY_NONE
-                SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER,
-                                [](int preverify, X509_STORE_CTX*) -> int
-                                {
-                                    return preverify;
-                                });
+                SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, [](int preverify, X509_STORE_CTX*) -> int {
+                    return preverify;
+                });
 
                 SSL_CTX_set_verify_depth(ctx, 4);
             }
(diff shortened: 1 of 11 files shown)


ixwebsocket/IXSocketOpenSSL.cpp
```diff
@@ -114,14 +114,17 @@ namespace ix
         SSL_CTX* ctx = SSL_CTX_new(_ssl_method);
         if (ctx)
         {
```
bsergean: shouldn't we have an else block here, that use SSL_VERIFY_NONE ? or is that not needed ?

ixwebsocket/IXSocketTLSOptions.h
```diff
@@ -0,0 +1,36 @@
...
+    {
+        // the certificate presented to peers
+        std::string cert_file;
+        // the key used for signing/encryption
+        std::string key_file;
+        // the ca certificate (or certificate bundle) file containing 
+        // certificates to be trusted by peers; use 'SYSTEM' to
+        // leverage the system defaults, use 'NONE' to disable peer verification
+        std::string ca_file = "SYSTEM";
+
```
bsergean: Can you put the body/implementation in a new file IXSocketTLSOptions.cpp please ?

ixwebsocket/IXSocketTLSOptions.h
```diff
@@ -0,0 +1,36 @@
...
+        bool isUsingClientCert() const {
+            return !cert_file.empty() && !key_file.empty();
+        }
+
+        bool isUsingSystemDefaults() const {
```
bsergean: indent nitpick: curly braces should be on their own line

ixwebsocket/IXWebSocket.cpp
```diff
@@ -73,6 +73,14 @@ namespace ix
         _perMessageDeflateOptions = perMessageDeflateOptions;
     }
 
+#ifdef IXWEBSOCKET_USE_TLS
```
bsergean: Let's get rid of the whole macro / ifdef stuff.

ixwebsocket/IXSocketTLSOptions.h
```diff
@@ -0,0 +1,36 @@
...
+ *  Copyright (c) 2017-2018 Machine Zone, Inc. All rights reserved.
+ */
+
+#pragma once
+
+#include <string>
+
+namespace ix
+{
+    struct SocketTLSOptions
```
bsergean: I have an idea, we can add an empty constructor which will just assert that IXWEBSOCKET_USE_TLS is on

#ifndef IXWEBSOCKET_USE_TLS
assert(false && "To use TLS feature the library must be compiled with USE_TLS);
#endif // something like that ?

ixwebsocket/IXSocketOpenSSL.cpp
```diff
@@ -283,17 +327,19 @@ namespace ix
             }
             SSL_set_fd(_ssl_connection, _sockfd);
```
bsergean: does SNI depends on peer validation ?

ixwebsocket/IXSocketOpenSSL.cpp
```diff
@@ -283,17 +327,19 @@ namespace ix
...
 
-            // SNI support
-            SSL_set_tlsext_host_name(_ssl_connection, host.c_str());
+            if (!_tlsOptions.isPeerVerifyDisabled()) {
```
bsergean: I said that somewhere else but curly braces should be on their own line.

ixwebsocket/IXSocketOpenSSL.cpp
```diff
@@ -265,12 +268,53 @@ namespace ix
...
+                    {
+                        unsigned long ssl_err = ERR_get_error();
+                        errMsg = "OpenSSL failed - SSL_CTX_default_verify_paths loading failed: ";
+                        errMsg += ERR_error_string(ssl_err, nullptr);
+                    }
+                } else {
+                    const char * root_ca_file = _tlsOptions.ca_file.c_str();
+                    STACK_OF(X509_NAME) *root_cas;
+                    root_cas = SSL_load_client_CA_file(root_ca_file);
+                    if (root_cas == NULL) {
```
bsergean: you could try: auto sslErr = ERR_get_error();

bsergean:
There is doc in the docs folder, would be great to add something there + adding parameters in ws command line tool + an example would be awesome (say providing a wrong cert and connecting to wss://echo.websockets.org or with ws curl).

matt-deboer:
ok; I'll get started on some of those.
btw -- do do you have an autoformatter that you use for this project in your own development? I'd love to just configure that and be done with it :)

ixwebsocket/IXSocketOpenSSL.cpp
matt-deboer: nope :) I was thinking of the security aspects, but the https->virtual host aspects would still apply; removed

ixwebsocket/IXSocketTLSOptions.h
matt-deboer: sounds good; did that

ixwebsocket/IXWebSocket.cpp
matt-deboer: yup

ixwebsocket/IXSocketOpenSSL.cpp
matt-deboer: I think that's default (0), but it's better to be clear; fixed

ixwebsocket/IXSocketTLSOptions.h
matt-deboer: done

ixwebsocket/IXSocketOpenSSL.cpp
matt-deboer: fixed

ixwebsocket/IXSocketOpenSSL.cpp
matt-deboer: ok

ixwebsocket/IXSocketTLSOptions.h
matt-deboer: yup

bsergean:
Thanks, just merged your work.
As a follow up it would be best to update the doc (see docs folder) and add support for those new options in the command line tools.

matt-deboer:
whoa, ok; I still had some follow up work on this--but I can just continue with another PR for that stuff

bsergean:
Sure. I merged a bit too soon indeed, the unites wasn't passing (I fixed and pushed it).

I need to have CI working fine again on GitHub.

Issue #118, opened 2019-10-23
Integer overflow in receiving code


bsergean:
(quoting the email exchange with a reporter)

There is an integer overflow on the receiving frame size https://github.com/machinezone/IXWebSocket/blob/0e9cf863cf57cfb02480302849b54dce5c0ac74a/ixwebsocket/IXWebSocketTransport.cpp#L545 that can be used to sneak past the guard clause _rxbuf.size() < ws.header_size + ws.N check.

Should be a simple fix - either by adding a reasonable limit to the ws.N value or checking for overflow before doing the addition. According to the websocket spec "the most significant bit MUST be 0" so that is a safe and simple check to do.

PR #119, opened 2019-10-26
check max frame size

bsergean:
With respect to ticket #118


bsergean:
A single WebSocket frame, per RFC-6455 base framing, has a maximum size limit of 2^63 bytes (9,223,372,036,854,775,807 bytes

bsergean:
Thanks, just merged it.

PR #122, opened 2019-11-20
Compile bug

fcojavmc:
1) IXWebSocketTransport: BUG: int type has no warranty of number of bits. It depends on compiler and architecture. In my system (64 bit) is 32 bit.

1 << 63 is bad idea in this case because the final number is 0 by overflow.
The symptom observed is that the server can't receive messages.

2) IXSocketFactory: Compilation Warning: Variable not in use.

### DIFF
--- a/ixwebsocket/IXSocketFactory.cpp
+++ b/ixwebsocket/IXSocketFactory.cpp
@@ -29,7 +29,11 @@ namespace ix
     std::shared_ptr<Socket> createSocket(bool tls,
                                          int fd,
                                          std::string& errorMsg,
+#ifdef IXWEBSOCKET_USE_TLS
                                          const SocketTLSOptions& tlsOptions)
+#else
+                                         const SocketTLSOptions&)
+#endif
     {
         errorMsg.clear();
         std::shared_ptr<Socket> socket;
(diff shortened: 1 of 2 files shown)


ixwebsocket/IXWebSocketTransport.cpp
```diff
@@ -543,7 +543,7 @@ namespace ix
             }
 
             // Prevent integer overflow in the next conditional
-            const uint64_t maxFrameSize(1 << 63);
+            const uint64_t maxFrameSize(uint64_t(1) << 63);
```
bsergean: Can you make this 1LL << 63 ?

ixwebsocket/IXSocketFactory.cpp
```diff
@@ -29,7 +29,11 @@ namespace ix
     std::shared_ptr<Socket> createSocket(bool tls,
                                          int fd,
                                          std::string& errorMsg,
+#ifdef IXWEBSOCKET_USE_TLS
                                          const SocketTLSOptions& tlsOptions)
```
bsergean: Can you try this trick, at the start of the function:

(void)tlsOptions;

ixwebsocket/IXWebSocketTransport.cpp
bsergean: or 1ULL or something like that, see if there's an option not to do the full cast to uint64_t ?

bsergean:
Thanks !

PR #136, opened 2019-12-30
Update IXSocketMbedTLS.cpp

mox1:
Doing more testing with mbedtls server mode on windows.
It was crashing upon startup, the actual error was that my private key file wasn't found.
Unfortunately, because some initialization was missing, it simply crashed. Traced the error back to the &_pkey structure not being initialized, crash started here (IXSocketMbedTLS.cpp, line 86)

`if (mbedtls_pk_parse_keyfile(&_pkey, _tlsOptions.keyFile.c_str(), "") < 0)`

This PR fixes the crash, it now correctly prints the mbedtls error message.

### DIFF
--- a/ixwebsocket/IXSocketMbedTLS.cpp
+++ b/ixwebsocket/IXSocketMbedTLS.cpp
@@ -48,6 +48,12 @@ namespace ix
 
         const char* pers = "IXSocketMbedTLS";
 
+        mbedtls_entropy_init(&_entropy);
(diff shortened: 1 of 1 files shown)


mox1:
Update: Looking over the code now, I see a slightly better place to put my updates. Closing this and will be sending an updated one shortly.

bsergean:
I was gonna comment about initMBedTLS, but it looks like you found that method already.

PR #139, opened 2019-12-31
Update IXSocketMbedTLS.cpp

mox1:
fix bug with mbedtls server certificate loading.

Prior to this PR, the certificate was only loaded if peer verification was enabled. 

Thus, if using a self-signed certificate and disabling cert verification meant that mbedtls never loaded the server cert. 

This fixes that issue, tested and working.


bsergean:
Excellent. Would you mind checking if the OpenSSL code has the same problem ?

mox1:
Was testing on Linux in server mode previously, did not encounter this problem. Also, looking at IXSocketOpenSSL.cpp it looks to be coded correctly.

PR #141, opened 2020-01-07
Fixed build error with vcpkg installed.

mattparks:
Including local headers to avoid confusion with possible Vcpkg installation.


bsergean:
Neat :) / I've been running in this issue while developing locally on my mac and switching SSL backends (mbedtls to openssl etc...). I would nuke the installed folders, but this isn't practical when using vcpkg.

Thanks !

PR #142, opened 2020-01-07
Allow configuration of Mbed TLS use.

mattparks:
I would like to use OpenSSL when building on Windows, currently `USE_MBED_TLS` is hard coded to TRUE on Windows. By using a option the value can be overridden. I also added a option for `USE_TLS`, using the current default value of FALSE.


bsergean:
I like that we add the ability to use openssl on windows, thanks for looking into that. Can you see my comment ?

bsergean:
I think I know what we should do:

```
if (WIN32 AND NOT USE_OPEN_SSL)
  set(USE_MBED_TLS TRUE)
endif()
```

bsergean:
You will have to handle this block too

```
if (USE_TLS)
    if (USE_MBED_TLS)
        list( APPEND IXWEBSOCKET_HEADERS ixwebsocket/IXSocketMbedTLS.h)
        list( APPEND IXWEBSOCKET_SOURCES ixwebsocket/IXSocketMbedTLS.cpp)
    elseif (APPLE AND NOT USE_OPEN_SSL)
        list( APPEND IXWEBSOCKET_HEADERS ixwebsocket/IXSocketAppleSSL.h)
        list( APPEND IXWEBSOCKET_SOURCES ixwebsocket/IXSocketAppleSSL.cpp)
    elseif (WIN32)
        list( APPEND IXWEBSOCKET_HEADERS ixwebsocket/IXSocketSChannel.h)
        list( APPEND IXWEBSOCKET_SOURCES ixwebsocket/IXSocketSChannel.cpp)
    else()
        set(USE_OPEN_SSL TRUE)
        list( APPEND IXWEBSOCKET_HEADERS ixwebsocket/IXSocketOpenSSL.h)
        list( APPEND IXWEBSOCKET_SOURCES ixwebsocket/IXSocketOpenSSL.cpp)
    endif()
endif()
```

I believe that 
```
    elseif (WIN32)
```

should become

```
    elseif (WIN32 AND NOT USE_OPEN_SSL)
```

Or we can move the openssl block to be right after the mbedtls one

mattparks:
I added those suggestions to my changes, I also moved the OpenSSL and Mbed TLS  options down into the first `USE_TLS ` block.

CMakeLists.txt
```diff
@@ -113,22 +113,27 @@ else()
...
     elseif (APPLE AND NOT USE_OPEN_SSL)
         list( APPEND IXWEBSOCKET_HEADERS ixwebsocket/IXSocketAppleSSL.h)
         list( APPEND IXWEBSOCKET_SOURCES ixwebsocket/IXSocketAppleSSL.cpp)
-    elseif (WIN32)
+    elseif (WIN32 AND NOT USE_OPEN_SSL)
         list( APPEND IXWEBSOCKET_HEADERS ixwebsocket/IXSocketSChannel.h)
         list( APPEND IXWEBSOCKET_SOURCES ixwebsocket/IXSocketSChannel.cpp)
-    else()
-        set(USE_OPEN_SSL TRUE)
+    elseif ()
```
bsergean: Why elseif with no condition instead of else ?

CMakeLists.txt
```diff
@@ -113,22 +113,27 @@ else()
...
 
-if (WIN32)
-  set(USE_MBED_TLS TRUE)
-endif()
+option(USE_TLS "Enable TLS support" FALSE)
 
 if (USE_TLS)
+    if (WIN32)
+        option(USE_MBED_TLS "Use Mbed TLS" ON)
+    else()
```
bsergean: I'm still unsure on how I can use mbed tls on macOS with that else case. Do you have a way to test that on linux or mac ?

This is the make target I use to test this.

```
ws_mbedtls:
        mkdir -p build && (cd build ; cmake -DCMAKE_BUILD_TYPE=Debug -DUSE_TLS=1 -DUSE_WS=1 -DUSE_MBED_TLS=1 .. ; make -j 4)
```

CMakeLists.txt
```diff
@@ -113,8 +113,11 @@ else()
     list( APPEND IXWEBSOCKET_HEADERS ixwebsocket/IXSelectInterruptEventFd.h)
 endif()
 
+option(USE_TLS "Enable TLS support" FALSE)
 if (WIN32)
-  set(USE_MBED_TLS TRUE)
+  option(USE_MBED_TLS "Use Mbed TLS" TRUE)
+else()
+  option(USE_MBED_TLS "Use Mbed TLS" FALSE)
```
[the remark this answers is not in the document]
mattparks: You would just need to pass `-DUSE_MBED_TLS=1` to your CMake generator args. If you are including IXWebSocket though `add_subdirectory` or `USE_MBED_TLS` these options can be configured with like this: 
```
set(USE_TLS ON CACHE INTERNAL "Add TLS support")
set(USE_MBED_TLS ON CACHE INTERNAL "Use Mbed TLS")
set(USE_OPEN_SSL OFF CACHE INTERNAL "Use OpenSSL")
```

CMakeLists.txt
mattparks: Should now be fixed in 61c9035021e1909f97a774254af87906b781ac95

PR #146, opened 2020-01-17
Doc update for Conan

LunarWatcher:
The PR to Conan Center has been merged \o/ 

This covers a tiny bit on the migration from the old Bintray remote, as well as the package references (it's version-dependent because of updates in 1.21.0).


bsergean:
Thanks !

Our company uses Android and I think that we could use conan to build third-party. We should look into it.

PR #147, opened 2020-01-26
Make WebsocketServer silent: Remove 'handleConnect() done' info message

ronen:
`IXWebsocketServer.cpp` has a `logInfo()` call, which reports when `handleConnect()` is finished.     There's no way to turn it off, so it makes the server noisy in normal use.

I'm not sure why it is in there, it's the only thing being logged.   So this PR simply removes it.

(Thanks for the nice library!)

### DIFF
--- a/ixwebsocket/IXWebSocketServer.cpp
+++ b/ixwebsocket/IXWebSocketServer.cpp
@@ -106,7 +106,6 @@ namespace ix
             }
         }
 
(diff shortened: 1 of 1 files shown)


bsergean:
Sure, I don't really see a problem with removing the log. What I should add is a way to enable logging with different level for the server code, so that those logs can be printed when debugging, but are off by default.

bsergean:
Thanks for the fix !

PR #149, opened 2020-01-28
bind returns zero on success

sethalves:


bsergean:
Thanks !

It looks like I messed that up in the previous commit, while working on the ipv6 support.

Issue #158, opened 2020-02-28
Linux build fails when zlib is not present on the system


rkrhlikar:
Building the project on Linux fails when zlib is not installed on the system. Running the `make` command attempts to build zlib from the third_party/zlib subdirectory. The process fails with the following errors (when building zlib examples):
```
[ 61%] Building C object third_party/zlib/CMakeFiles/example.dir/test/example.o
.../IXWebSocket/third_party/zlib/test/example.c:8:10: fatal error: zlib.h: No such file or directory    
 #include "zlib.h"
          ^~~~~~~~
```
```
[ 63%] Building C object third_party/zlib/CMakeFiles/minigzip.dir/test/minigzip.o
.../IXWebSocket/third_party/zlib/test/minigzip.c:18:10: fatal error: zlib.h: No such file or directory
 #include "zlib.h"
          ^~~~~~~~
```

This is a known issue with zlibs CMakeLists.txt (as reported [here](https://github.com/madler/zlib/issues/218)).

rkrhlikar:
As the zlib project seem to be inactive, the linked issue is unlikely to receive a fix. I opened a pull request (#159) that uses a workaround to resolve the problem with a misconfigured/missing include path in zlibs CMakeLists.

PR #159, opened 2020-02-28
Move zlib include_directories before add_subdirectory

rkrhlikar:
Moving the `include_directories` command before the `add_subdirectory` adds the listed include directories to the list used by the CMake project in the subdirectory.

This resolves issue #158


bsergean:
Thank you !

Issue #165, opened 2020-03-17
IXWebSocket will timeout 100% of the time if the ping interval and ping timeout are the same amount of time.


maxweisel:
Running the following will cause a connection close after five seconds. I believe it's because the first ping is sent out after five seconds and the first timeout check also happens at five seconds. In practice, I imagine the ping should be sent immediately on connect so it has time to return before the first timeout check.

```
WebSocket webSocket;
webSocket.setPingInterval(5);
webSocket.setPingTimeout(5);
webSocket->setUrl("<URL>");
webSocket->start();
```

I think I can fix this one easily. I'm going to try to write up a PR real quick.

Max

maxweisel:
Opened a PR here: https://github.com/machinezone/IXWebSocket/pull/166

bsergean:
Thanks, looking at it now.

bsergean:
Should be fixed in 8.3

PR #166, opened 2020-03-17
Fix bug where setting a ping interval and ping timeout of the same length will cause the socket to timeout immediately.

maxweisel:
Set the last send ping time point to _pingIntervalSecs earlier so it fires immediately upon connect. Otherwise if the timeout time is the same, it will not have time to return before the timeout check is run.

### DIFF
--- a/ixwebsocket/IXWebSocketTransport.cpp
+++ b/ixwebsocket/IXWebSocketTransport.cpp
@@ -235,7 +235,7 @@ namespace ix
     {
         {
             std::lock_guard<std::mutex> lock(_lastSendPingTimePointMutex);
-            _lastSendPingTimePoint = std::chrono::steady_clock::now();
+            _lastSendPingTimePoint = std::chrono::steady_clock::now() - std::chrono::seconds(_pingIntervalSecs);
         }
         {
             std::lock_guard<std::mutex> lock(_lastReceivePongTimePointMutex);


maxweisel:
This fixes issue #165

bsergean:
The bug report is legit but I don't think that the fix is correct. Those timing should be absolute, not relative.

Maybe the API should be changed so that we supply a multiplier of the ping-interval (like 2x or 3x). I usually do that. (send ping every 30s, and fail the connection if nothing has been received for 90s).

We should look at how other libraries do this (websockets python), to validate parameters.

maxweisel:
How do you mean? Those timings are absolute. The only change is that the `_lastSendPingTimePoint` is initialized to `_pingIntervalSecs` seconds before `now()` instead of being initialized to `now()` in order to prime the connection that it needs to send a ping immediately on connect.

The alternative fix could be to set the `_lastReceivePongTimePoint` to `now()` when the first ping is sent, but that would essentially mean that the ping timeout duration would be `(_pingTimeoutSecs + _pingIntervalSecs)` seconds when the connection is initially made before returning back to `_pingTimeoutSecs` seconds worth of time after the first pong is received.

bsergean:
Ok I think I understand better your fix now. You're right time are already absolute.

1. Can you add a tiny extra timing, so that we are sure that the time will be exceeded for sure, regardless of rounding errors (say one second)

_lastSendPingTimePoint = std::chrono::steady_clock::now() - std::chrono::seconds(_pingIntervalSecs + 1);
2. Can you add a comment explaining briefly the problem, and why we need to initialize the time point to something 'prior' to now ? 

Thanks

maxweisel:
A few quick notes:

1. Yes definitely

2. I just deployed this fix to production and this fix can still cause problems if the client and server are running on the same machine (explained below). Before I get into that, here's an outline of the original issue. For simplicity, let's say now() is measured in seconds

```
/* Initialization occurs */
webSocket.setPingInterval(5);
webSocket.setPingTimeout(5);

/* Socket Connect event fires (now() == 100) */
_lastSendPingTimePoint    = now(); // 100
_lastReceivePongTimePoint = now(); // 100

/* time passes */

/* WebSocketTransport::poll() runs (now() == 106) */

// Check pingTimeoutExceeded()
if (now() - _lastReceivePongTimePoint > _pingTimeoutSecs) { //  (106 - 100 > 5)
    // This block is entered and the connection is closed. We have not sent any ping messages yet, so it's expected we wouldn't have received any pong messages either.
}

// Check pingIntervalExceeded()
// - Doesn't matter because the ping timeout already fired.

```

The issue above is that `pingTimeoutExceeded()` evaluates to true in `poll()` before we ever send any pings. However, even if we swapped the order of pingIntervalExceeded() and pingTimeoutExceeded(), the ping wouldn't have enough time to get a pong back and we'd still fail.

My original idea was to set `_lastSendPingTimePoint` to a value equal to `now() - _pingTimeoutSecs` so that way `pingIntervalExceeded()` evaluates to `true` the first time `poll()` is executed.

This solution does work in the majority of cases. The ping goes out, and a pong comes back and sets `_lastReceivePongTimePoint` to a later time, which allows the next ping to go out before we timeout, and the process repeats forever.

However, if both the client and server are on the same machine, the initial ping goes out, and it comes back before the next `poll()`. The result is that `_lastReceivePongTimePoint` is set to a timestamp such that the timeout will be exceeded on the same `poll()` frame that we plan to send the next ping. This one is slightly harder to fix because we're essentially dealing with a race condition at this point. And even if we fix this case, it's possible the two timestamps will become close enough together to trigger this bug again later on in the connection.

Personally, the way I would fix this is to ditch the two separate timestamps and have a single timestamp for the last send time, and a boolean for whether we've received a pong since the last send. The code would work like this:

```
void ConnectionOpen() {
    // Send a ping on connection open
    _hasReceivedPong = false;
    sendPing();
}

void CheckAndSendPing() {
    // Check if it's time to send another ping
    if (now() - _lastSendPingTimePoint < _pingIntervalSecs)
        return;

    // It's time to send a ping, let's make sure we got a pong back first
    if (!_hasReceivedPong) {
        close(<timeout>);
        return;
    }

    // We're good! Reset and send another ping
    _hasReceivedPong = false;
    sendPing();
}

void sendPing() {
    send(PING);
    _lastSendPingTimePoint = now();
}

This would solve both the original issue, and the instantaneous pong issue too. If WebSockets were over UDP, I'd understand the need to possibly wait for more than one pong, but over TCP, it doesn't make sense to send them more often than your timeout.

Max

bsergean:
You're right that it would be much simpler to implement it like you did, and get rid of setPingTimeout altogether. I've never liked too much the fact that we have 2 things to configure. Could you check how the 'ws' javascript library do it, I think they also only have one timeout ?

We still need a bit of complexity for computing the timeout we pass to poll, but going from 2 timepoints to 1 is a win.

If you want to make a PR for it that would be appreciated, otherwise I'll get to it at some point.

The only downside with this approach is that 'missing' a pong is lethal.

bsergean:
I'm going to try to implement it fyi.

bsergean:
I just implemented something in master (8.3), please give it a try when you get a chance.

bsergean:
I think we're better off with the other approach you suggested and I implemented in 8.3.

maxweisel:
My apologies for the delay! Will give this a shot next week. Thanks for writing up a fix!

Max

maxweisel:
For what it's worth, I don't believe the javascript ws package provides any ping-based heartbeat out of the box. I believe it's something developers need to implement themselves. That said, I really love that this library provides it and also automatic reconnect with exponential backoff. I've had to implement those myself in node and it's always a little unnerving to reinvent the wheel for a foundational element like that.

maxweisel:
Yeah, it looks like they ask developers to implement it themselves: https://github.com/websockets/ws#how-to-detect-and-close-broken-connections

The good news is the recommended implementation matches the patch you've implemented.

bsergean:
Great. Well tell me how it goes. Thanks for looking at what 'ws' is doing, I wasn't aware. I'm using a python library which implements it.

For us it worked ok without an explicit ping/pong until we used a service running on AWS, and with the Amazon Load Balancer it was impossible to detect that a remote end had been closed at the socket level. I believe that before that we used a hardware based load balancer (F5) and it was ok.

And be well in this crazy corona-virus world, I'm up early today because I had to go in line early at Safeway to buy some toilet-paper before they run out :p !!

PR #169, opened 2020-03-26
Only find zlib if it has not already been found.

mattparks:


bsergean:
Hey can you elaborate a bit on how you think this should be used ?

Are you setting ZLIB_FOUND in your own CMakefile, because you already have a copy around ?
I tried to find example of usage but it's a bit wild.

Your change seems safe, but what I don't want to break is vckpg code.
Also we could add variables like 'use bundled zlib' or equivalent. Was looking at libwebsockets and this is what they do.

mattparks:
Yes I'm setting ZLIB_FOUND in my own project. I could
set CMAKE_DISABLE_FIND_PACKAGE_ZLIB to true, but this would disable finding
zlib completely.
This change shouldn't break existing usages, zlib will not be already found
in most cases. This type of check is already done around OpenSSLs
find_package.

bsergean:
Thanks !

The CMakelist code is a bit convoluted, if you can find a way to make it more consistent (maybe with a new USE_BUNDLED_ZLIB, or something along those line) I'd appreciate it.

Is your project open-source or closed source ?

mattparks:
USE_BUNDLED_ZLIB would be a good option to check for if zlib is not found
at all. My project is closed-source.

PR #178, opened 2020-04-13
Implement API for adding custom roots via a string

maxweisel:
Heyo,

I don't think this is ready to merge, but it's at a point where I'd like some feedback on the API design. I feel good about the implementation in IXSocketOpenSSL, but there isn't really a clean way to specify a string for root CAs via SocketTLSOptions.

I've implemented a quick API on SocketTLSOptions so I can test, but I don't think it's the right way to specify a string of certs. Let me know what you think is the ideal way to specify a string cert in SocketTLSOptions and I can implement and update this PR.

Max

### DIFF
--- a/ixwebsocket/IXSocketMbedTLS.cpp
+++ b/ixwebsocket/IXSocketMbedTLS.cpp
@@ -103,10 +103,18 @@ namespace ix
             {
                 ; // FIXME
             }
-            else if (mbedtls_x509_crt_parse_file(&_cacert, _tlsOptions.caFile.c_str()) < 0)
-            {
-                errMsg = "Cannot parse CA file '" + _tlsOptions.caFile + "'";
-                return false;
+            else {
+                if (_tlsOptions.isUsingInMemoryCAs()) {
+                    const char *buffer = _tlsOptions.caFile.c_str();
+                    size_t      bufferSize = _tlsOptions.caFile.size() + 1; // Needs to include null terminating character otherwise mbedtls will fail.
+                    if (mbedtls_x509_crt_parse(&_cacert, (const unsigned char *)buffer, bufferSize) < 0) {
+                        errMsg = "Cannot parse CA from memory.";
+                        return false;
+                    }
+                } else if (mbedtls_x509_crt_parse_file(&_cacert, _tlsOptions.caFile.c_str()) < 0) {
+                    errMsg = "Cannot parse CA file '" + _tlsOptions.caFile + "'";
+                    return false;
+                }
             }
 
             mbedtls_ssl_conf_ca_chain(&_conf, &_cacert, NULL);
--- a/ixwebsocket/IXSocketOpenSSL.cpp
+++ b/ixwebsocket/IXSocketOpenSSL.cpp
@@ -460,20 +460,22 @@ namespace ix
                 if (_tlsOptions.isUsingInMemoryCAs()) {
                     // Load from memory
                     openSSLAddCARootsFromString(_tlsOptions.caFile);
-                } else if (SSL_CTX_load_verify_locations(
+                } else {
+                    if (SSL_CTX_load_verify_locations(
                              _ssl_context, _tlsOptions.caFile.c_str(), NULL) != 1)
-                {
-                    auto sslErr = ERR_get_error();
-                    errMsg = "OpenSSL failed - SSL_CTX_load_verify_locations(\"" + _tlsOptions.caFile +
-                             "\") failed: ";
-                    errMsg += ERR_error_string(sslErr, nullptr);
-                    return false;
+                    {
+                        auto sslErr = ERR_get_error();
+                        errMsg = "OpenSSL failed - SSL_CTX_load_verify_locations(\"" + _tlsOptions.caFile +
+                                 "\") failed: ";
+                        errMsg += ERR_error_string(sslErr, nullptr);
+                        return false;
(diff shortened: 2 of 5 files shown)


maxweisel:
For what it's worth, I think the ideal implementation would add an enum to SocketTLSOptions that lets you specify the certificate type (system, none, file, string). I don't think using caFile as both a path and a type is the right move, but changing that will break all projects that use SocketTLSOptions so I'm unsure the best way to move forward.

Max

bsergean:
Thanks for the contribution.

An enum would have been nice indeed but it's a bit late.

There is a tls boolean in the SocketTLSOptions struct. If we add a 'memory' boolean (as opposed to file), set to false, we can use that to drive the new behavior and still maintains backward compatibility.

We will need to expose this to the ws tool, and to have one unitest that covers this new code path (and the old ones too). At least for the client code. Can you look at how this would work for mbedlts ?

maxweisel:
So what do you think about this new implementation. Essentially I check the caFile string for `-----BEGIN CERTIFICATE-----`. If that's present, then I treat it as an in-memory string of certificates, otherwise it falls back to the previous behavior of loading as a file from disk.

Once the API design feels right, I can go ahead and implement the mbedtls / Security.framework versions.

Max

maxweisel:
Just added the mbedtls implementation too.

bsergean:
The BEGIN CHECK feels a bit hacky.

We could prefix the memory, or the path with a url like scheme, and when there's nothing we'd still use the 

file:///usr/lib/openssl/cacert.pem <file:///usr/lib/openssl/cacert.pem>

memory://----BEGIN <memory://----BEGIN> CERTIFICATE ....

We could have utilities to set string that add those prefix.

There's DER format too.

Last thing is that the OS X implementation might be tough.

maxweisel:
I was considering the schema approach, but adding file:/// would break projects that expect the old API right? I don't believe anyone is currently submitting file:/// paths. afaik the library doesn't currently handle them properly. If we're up for introducing breaking changes, I'd love to go back to the enum approach. The check for -----BEGIN CERTIFICATE----- was the closest method I could think of style-wise to checking for NONE and SYSTEM.

As far as DER goes, I don't think there's a way to add DER support without breaking backwards compatibility. I don't think we can safely use std::string for anything binary encoded as we can't guarantee it won't go through UTF encoding/decoding when moving in/out of being a binary format. It's also worth noting that DER does not support more than one certificate, so we'd probably want to provide an API that lets people specify an array of DER encoded certificates rather than using an std::string.

As far as the Security framework goes, I've had success using this API in the past: https://developer.apple.com/documentation/security/1392400-sslsetcertificate?language=objc Is there a reason that wouldn't work here?

Max

bsergean:
> On Apr 14, 2020, at 2:45 PM, Max Weisel <notifications@github.com> wrote:
> 
> 
> I was considering the schema approach, but adding file:/// would break projects that expect the old API right? I don't believe anyone is currently submitting file:/// paths. afaik the library doesn't currently handle them properly. If we're up for introducing breaking changes, I'd love to go back to the enum approach. The check for -----BEGIN CERTIFICATE----- was the closest method I could think of style-wise to checking for NONE and SYSTEM.
> 
We would treat legacy values thing as file path. If there's an extra file:// or memory:// we would act accordingly.

> As far as DER goes, I don't think there's a way to add DER support without breaking backwards compatibility. I don't think we can safely use std::string for anything binary encoded as we can't guarantee it won't go through UTF encoding/decoding when moving in/out of being a binary format. It's also worth noting that DER does not support more than one certificate, so we'd probably want to provide an API that lets people specify an array of DER encoded certificates rather than using an std::string.
> 
Ok let's forget about DER then.
> As far as the Security framework goes, I've had success using this API in the past: https://developer.apple.com/documentation/security/1392400-sslsetcertificate?language=objc <https://developer.apple.com/documentation/security/1392400-sslsetcertificate?language=objc> Is there a reason that wouldn't work here?
> 

Give it a try, it was just a hint. I tried to do what curl is doing once without too much success, but I was also working on the server side. Maybe handling just the client side is easier.

bsergean:
Can you run this:

```
$ make format 
clang-format -i `find test ixwebsocket ws -name '*.cpp' -o -name '*.h'`
```

and put curly braces on their own line if that doesn't do it. I'd like to keep the style consistent.

bsergean:
Can you try the instructions in https://machinezone.github.io/IXWebSocket/ws/#cobra-client-and-server?

But instead of passing in a file for cert etc... use cat and double quote to pass the cert on the command line. 

```
--cert-file "`cat ${certs}/trusted-client-crt.pem`"
```

bsergean:
2 extra thoughts:

1. std::string works ok with binary, this is what we use internally in this library to process websocket binary type messages. std::string are not \0 terminated
2. I don't think that osx secure-transport support PEM, so the string search for '--BEGIN CERTIFICATE' is unlikely to work there.
3. It might be ok to just do the BEGIN CERTIFICATE thing to keep the ball rolling.

bsergean:
"good enough" :)

I reformatted your code to match the existing style.

maxweisel:
Apologies for the delay! I've been slammed with another project. Thanks for reformatting and merging!

Max

bsergean:
No worries.

ps: I wonder if some the new loading from memory code could be used for cert and 'normal' key loading.

Issue #155, opened 2020-02-19
HTTP Client: Payload to short


Naruk-TUBK:
With HTTP Requests i always get to short response payloads for GET and PUT requests. The payload is about 1/3 of the expected payload. It is no server Problem because with libcurl it works.
This is my c++ code. Is something wrong with it?

`ix::HttpClient httpClient;
            ix::HttpRequestArgsPtr args = httpClient.createRequest();
            ix::WebSocketHttpHeaders headers;
            headers["Content-Type"] = "application/json";
            args->extraHeaders = headers;
            // Timeout options in seconds
            args->connectTimeout = 4;
            args->transferTimeout = 4;
            ix::HttpResponsePtr out;
            out = httpClient.get("http://localhost:8181/uri", args);
            std::cout << out->payload << std::endl;`

bsergean:
Thanks for reporting this.

What is the server code ?
Could you write a unittest for it, or a minimal server ?
Can you run this against nginx for example, in the GET case ?

I reformatted your code comment:

```
ix::HttpClient httpClient; 
ix::HttpRequestArgsPtr args = httpClient.createRequest(); 
ix::WebSocketHttpHeaders headers; 
headers["Content-Type"] = "application/json"; 
args->extraHeaders = headers; // Timeout options in seconds 
args->connectTimeout = 4; 
args->transferTimeout = 4; 
ix::HttpResponsePtr out; 
out = httpClient.get("http://localhost:8181/uri", args); 
std::cout << out->payload << std::endl;
```

Naruk-TUBK:
I want to access a smart home api from https://github.com/dresden-elektronik/deconz-rest-plugin. I just noticed that i encounter the problem only with there api, therefor it is hard for me to write a test. But i just printed out the response headers with:

    for(auto& header : out->headers) {
        std::cout << header.first << " " << header.second  << "\n";
     }

Header with IXWebsockets:

    Access-Control-Allow-Origin *
    Content-Encoding 
    Content-Length 906
    Content-Type application/json; charset=utf-8
    ETag e545399fd0d2c0a89ceb72c669380d87"

Header with curl (curl http://localhost:8181/uri -i)
     HTTP/1.1 200 OK
    Access-Control-Allow-Origin: *
    Content-Type: application/json; charset=utf-8
    Content-Length:4906
    ETag:"e545399fd0d2c0a89ceb72c669380d87"

[...]

PR #184, opened 2020-04-25
Bug on setting extra headers. Now it loses the first string character.

fcojavmc:
Client code:

    ...

    ix::WebSocketHttpHeaders headers {
        {"Cookie", "ABC"}
    };

    ...

Expected header string on server:
    "Cookie: ABC"

Resulted header string on server:
    "Cookie: BC"

Solution:
    The easy way I found to solve the problem is to add a space where extra headers are set before sended to server.

### DIFF
--- a/ixwebsocket/IXWebSocketHandshake.cpp
+++ b/ixwebsocket/IXWebSocketHandshake.cpp
@@ -133,7 +133,7 @@ namespace ix
 
         for (auto& it : extraHeaders)
         {
-            ss << it.first << ":" << it.second << "\r\n";
+            ss << it.first << ": " << it.second << "\r\n";
         }
 
         if (_enablePerMessageDeflate)


bsergean:
Thanks for the bug fix and the report. 

Let me think about the best way to fix that, I'll see if I can add a unittest too. There was a header related change that went in, maybe it's a regression of that.

See #155

bsergean:
I think I know what's going, your server might not be compliant, as it expects a space after the header name, and the http spec does not requires this.

But there's no harm in adding an extra space, this is how it used to be. I had just removed that space for testing.

bsergean:
Merged, doc'ed it and bumped the version number.

Muchas gracias from San Francisco, Francisco :)

fcojavmc:
Thank you, Benjamin.

The server I used was IXWebSocket server. I thought that the problem was in IXWebSocketHttpHeaders.cpp arround line 65, but seems ok. Please fix-it if you have time to figure out where's the problem when the client do not put a space after colon. Javascript webSocket use a space but it's not mandatory.

Best regards from Mallorca, Spain ;)

bsergean:
Ok I have merged your change, tell me if it works for you now.

Say hi to Rafael Nadal from me :)

bsergean:
ps: out of curiosity, what platform/OS are you on ?

fcojavmc:
Hello, Benjamin.

I'm using GNU/Linux for programming, compiling and cross compiling and I'm working on projects for Linux, Windows and Android.

I like your server because low external dependencies and using modern C++. I have legacy projects that need to still work on Windows XP but your server is Windows Vista minimum required version. I'll try to do something with that but it's not really urgent because I'm using Qt WebSocket server for now. (Qt 5.6 still works on Windows XP).

Good job, Benjamin!

bsergean:
Cool that’s interesting.

Do you know what it would take to support Windows XP ?
Have you tried compiling on this platform ? 

I don’t think that it would be too hard to make the library C++11, instead of C++14.

Issue #191, opened 2020-05-03
CMake export regression


bsergean:
cf #190

@ebenali do you know what we can do about this windows error ?

https://pipelines.actions.githubusercontent.com/SYUiO8aHxb4LIyZgymGkvgwhJRrl1FTnKw8Q829qRwtlbSK3Tg/_apis/pipelines/1/runs/273/signedlogcontent/12?urlExpires=2020-05-03T03%3A59%3A43.8936693Z&urlSigningMethod=HMACV1&urlSignature=FPNAHaDaUlpuVdo6oibovpRGGUVFcg7ElapeXiT5rEM%3D

ebenali:
pull-req #192 should fix this issue.

bsergean:
I think it did. Thanks @ebenali !

PR #192, opened 2020-05-03
Fix CMake/zlibstatic-related regression

ebenali:
Last commit fixes regression #191  involving pulling in zlib static library.


bsergean:
Thanks !

PR #193, opened 2020-05-03
Added asynchronous udp receive function

tostc:
Hi,

I've added a new function for the UPDSocket. It allows to receive asynchronous data over udp.

### DIFF
--- a/ixwebsocket/IXUdpSocket.cpp
+++ b/ixwebsocket/IXUdpSocket.cpp
@@ -9,7 +9,6 @@
 #include "IXNetSystem.h"
 #include <cstring>
 #include <sstream>
-#include <fcntl.h>
 
 namespace ix
 {
--- a/ixwebsocket/IXUdpSocket.cpp
+++ b/ixwebsocket/IXUdpSocket.cpp
@@ -10,6 +10,7 @@
 #include <cstring>
 #include <sstream>
 #include <thread>
+#include <fcntl.h>
 
 namespace ix
 {
@@ -45,6 +46,18 @@ namespace ix
         return err;
     }
 
+    bool UdpSocket::isWaitNeeded()
+    {
+        int err = getErrno();
+
+        if (err == EWOULDBLOCK || err == EAGAIN || err == EINPROGRESS)
+        {
+            return true;
+        }
+
+        return false;
+    }
+
     void UdpSocket::closeSocket(int fd)
     {
 #ifdef _WIN32
@@ -62,6 +75,13 @@ namespace ix
             errMsg = "Could not create socket";
             return false;
         }
+
+#ifdef _WIN32
+        unsigned long nonblocking = 1;
+        ioctlsocket(sockfd, FIONBIO, &nonblocking);
+#else
+        fcntl(_sockfd, F_SETFL, O_NONBLOCK); // make socket non blocking
+#endif
 
         memset(&_server, 0, sizeof(_server));
         _server.sin_family = AF_INET;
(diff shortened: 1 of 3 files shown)


bsergean:
Thanks for contributing @tostc I appreciate it. Are you implementing a standard protocol with this UDP code ? What is it that you are trying to do now ?

In its current form I'm not sure I want to merge this ; if you look at the other socket class (IXSocket, which is TCP), it does not have a thread associated. They are low level and not threaded. The socket can be used in non-blocking mode btw.

If you really like your design, I would suggest you make a derived class from it to implement the threaded async way.

It is quite inefficient to start a thread for every message received. Same deal for some buffer, which could be member variables.

I noticed you added code to 'receive' data, which was missing. If you want to add just that to the library I would merge it. The signature should be like that: `ssize_t UdpSocket::recv(void* buffer, size_t length)`

tostc:
I will remove the async method and add a low level, non-blocking version of receive. I don't think that the protocol which I implement is a standard one. It's the ip discovery protocol of Discord. If you are interested in the protocol look [here](https://discordapp.com/developers/docs/topics/voice-connections#ip-discovery). It's the link to the Discord API documentation.

tostc:
Ok I've changed the method.

ixwebsocket/IXUdpSocket.cpp
```diff
@@ -9,6 +9,7 @@
 #include "IXNetSystem.h"
 #include <cstring>
 #include <sstream>
```
bsergean, line 11: I have a feeling that this include will blow up on Windows. Can you move it to the UNIX portion of IXNetSystem.h ?

ixwebsocket/IXUdpSocket.cpp
```diff
@@ -93,4 +113,11 @@ namespace ix
         return (ssize_t)::sendto(
             _sockfd, buffer.data(), buffer.size(), 0, (struct sockaddr*) &_server, sizeof(_server));
     }
+
+    ssize_t UdpSocket::recvfrom(void* buffer, size_t length)
+    {
```
bsergean, line 118: Can you put the sizeof(_server) directly in the function call ?
I am afraid of compiler warnings.

bsergean:
Just looked at the discord api, that's cool.

1. Are you aware of this ? https://gitlab.com/HCInk/dis-light
2. Can I add a reference to your project as a user of IXWebSocket on our README page ?

liz3:
dis-light does indeed implement audio sending. but i implemented that using the low level apis by the os recvfrom/sendto: https://gitlab.com/HCInk/dis-light/-/blob/master/lib/voice/voice_connection.cpp

The ip discovery is handled here: https://gitlab.com/HCInk/dis-light/-/blob/master/lib/voice/voice_connection.cpp#L189-212

tostc:
> Just looked at the discord api, that's cool.
> 
>     1. Are you aware of this ? https://gitlab.com/HCInk/dis-light
> 
>     2. Can I add a reference to your project as a user of IXWebSocket on our README page ?

1. I've first heard of it after I've opened this pull request :smile: 
2. Of course, but the library is in very early development stage. If you want to add it anyhow you can find the library [here](https://github.com/tostc/libDiscordBot/tree/master).

bsergean:
@tostc are you able to use the merged code ?
I had a bunch of trouble to get the code to compile on Windows and Unix but it ultimately did.

tostc:
I'm so sorry for the trouble. I've tested my changes only on Linux and it work's fine. I've checked only the Windos docs, if the function exists. I don't have checked the parameter types.

bsergean:
no worries.

PR #195, opened 2020-05-05
add PATCH and add option to enforce a http request body write

liz3:
Discord uses http PATCH for things like message edit which require a body. so i added patch.
Further i know apis also allowing/requiring bodies on other methods such as DELETE, but rather then just pushing a body on every method. i though adding a property to enforce it would be a better approach.

### DIFF
--- a/ixwebsocket/IXHttpClient.cpp
+++ b/ixwebsocket/IXHttpClient.cpp
@@ -25,6 +25,7 @@ namespace ix
     const std::string HttpClient::kHead = "HEAD";
     const std::string HttpClient::kDel = "DEL";
     const std::string HttpClient::kPut = "PUT";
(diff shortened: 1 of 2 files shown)


bsergean:
Hey thanks for the 'patch' (pun intended).

I think I would prefer for consistency that we do exactly what I did for put and post, this way all the code looks the same.

liz3:
@bsergean isnt that what i did with patch?

liz3:
The issue the boolean addresses the cases where it might happen that certain unexpected methods require bodies. there should be a way to include a body unrelated to the request method

bsergean:
Ok, I merged and I reworked the patch a bit -> 1535518

I didn't know there was a use case for forcing a body, I don't think the new method hurts.

PR #197, opened 2020-05-08
Move the fPIC option to the top of the CMakeLists

XLPhere:
Move the fPIC-option (`CMAKE_POSITION_INDEPENDENT_CODE`) to the top of the CMakeLists, otherwise it will still give the following error when linking against `libixwebsocket.a` (with `-lixwebsocket`) on linux (gcc):
> `/usr/bin/ld: /usr/local/lib/libixwebsocket.a(IXHttpClient.cpp.o): relocation R_X86_64_PC32 against symbol '_ZN2ix10HttpClient3runEv' can not be used when making a shared object; recompile with -fPIC`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,10 @@ set (CMAKE_CXX_STANDARD 14)
 set (CXX_STANDARD_REQUIRED ON)
 set (CMAKE_CXX_EXTENSIONS OFF)
 
+if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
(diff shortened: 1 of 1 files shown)


bsergean:
Thanks !
@liz3 have you seen this ?

I have seen the use of fpic through cmake heavily in other projects so it looks like the right thing to do on Linux when using IXWebsocket in a shared library.

liz3:
@bsergean yes.
If it does not add any other unwanted behaviour or issues then it should be safe to use and well it does not work without. so ye.

PR #201, opened 2020-05-15
Update CMakeLists.txt

flagarde:
just have to provide OPENSSL_ROOT_DIR to cmake


bsergean:
There's been tons of changes in that area, and that might break others who did similar changes.

Way back there was just a find_package here. Maybe we keep the find_package inside a if statement.

@mattparks + @ebenali + @rkrhlikar any opinion ?

bsergean:
@flagarde I merged your code in 789e6204515999c08ebe2aaccc90a5dceccd0ca2 / and added a note to OPENSSL_ROOT_DIR

PR #204, opened 2020-05-18
Extend vcpkg docs

liz3:
Adds linking instructions for cmake when using vcpkg

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -42,6 +42,19 @@ It is possible to get IXWebSocket through Microsoft [vcpkg](https://github.com/m
 ```
 vcpkg install ixwebsocket
 ```
+To use the installed package within a cmake project, use the following:
+```cmake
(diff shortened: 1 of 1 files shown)


bsergean:
9.6.2 is available on vcpkg now fyi.

liz3:
@bsergean yes, i updated my vcpkg installation and it also uses 9.6.2.

But even after specifying openssl it seams its still including mbdedtls.
Do i need to explicitly exclude mbded tls? 
image

bsergean:
... so I must have messed up something, but I'm afraid I've reached my vcpkg knowledge limits.

1. Can you ask over there ? Maybe they have an FAQ or a forum or something
2. Is mbedtls not an option for you ?

liz3:
I can ask them. 
As soon as i do **not** need to specify a custom root cert(like it was the last time/currently?) mbedtls is fine.

bsergean:
I have made it so that the system cert is loaded by default with mbedtls. So it should be easy to use and equivalent to openssl, in ea207d81992a9ef91cfc50fc6902ae4fdb705af0

Specifying a custom root cert should work equally well with both backends.

liz3:
@bsergean YES. works fine. thanks a lot for the hard. work!!
image

PR #206, opened 2020-05-20
Re-enable support for clang 5 and earlier

LunarWatcher:
A change introducing std::atoi to IXUrlParser broke support for clang 5 and earlier. See conan-io/conan-center-index#1688. This can be fixed by including `<cstdlib>`. `<cstring>` only contains the relevant functions with Clang 6 and up. Feel free to mess around with the compiler version on Godbolt. The first link (Clang 5, no cstdlib) starts compiling on Clang 6 and up.

Clang 5, no cstdlib: https://godbolt.org/z/9pNAd6
Clang 5, cstdlib: https://godbolt.org/z/qSNt-Z

### DIFF
--- a/ixwebsocket/IXUrlParser.cpp
+++ b/ixwebsocket/IXUrlParser.cpp
@@ -33,6 +33,7 @@
 
 #include <algorithm>
 #include <cstring>
+#include <cstdlib>
 
 namespace
 {


bsergean:
Thanks !

bsergean:
I just released 9.6.4 for your fix.

PR #207, opened 2020-05-21
Update CMakeLists.txt

flagarde:
Using ZLIB_ROOT as described in https://cmake.org/cmake/help/v3.4/module/FindZLIB.html appears not to work because of the policy CMP0074

setting  cmake_minimum_required(VERSION 3.4.1...3.17.2) fix the issue

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@
 # Copyright (c) 2018 Machine Zone, Inc. All rights reserved.
 #
 
-cmake_minimum_required(VERSION 3.4.1)
+cmake_minimum_required(VERSION 3.4.1...3.17.2)
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMake;${CMAKE_MODULE_PATH}")
 
 project(ixwebsocket C CXX)


bsergean:
Can you explain why does the ... mean between the 2 versions ?

We still want to support users on an old cmake, maybe we can bump the cmake min version but going from 3.4 all the way to 3.17 is a big jump ...

bsergean:
If you use cmake 3.17.2, does FindZLIB work for you ?

flagarde:
Hi I think the ... pattern is exactly what you are looking for.
From the cmake doc : https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html
If the running version of CMake is lower than the <min> required version it will stop processing the project and report an error. The optional <max> version, if specified, must be at least the <min> version and affects policy settings as described below. If the running version of CMake is older than 3.12, the extra ... dots will be seen as version component separators, resulting in the ...<max> part being ignored and preserving the pre-3.12 behavior of basing policies on <min>.

So if I understand correctly user using version 3.4.1 wouldn't see any change but user with a more up to date version should have their policies updated.

In 3.17.2 the warning disappears and the compilation is successful. But I didn't check all the options.

bsergean:
Thanks for the clear explanation. Let's give it a try.

PR #213, opened 2020-06-13
feat: add cmake find config support

xizhibei:
Hi, I found your lib quite useful, but it did not support cmake `find_package`, and it was not convenient to use in my project.

so I add the cmake find config support, so now we can just use your lib like this:

```cmake
find_package(ixwebsocket REQUIRED)
target_link_libraries(mylib PUBLIC ixwebsocket::ixwebsocket)
```


bsergean:
Code change looks good, but there was a similar change that created some problem for one user who's trying to link with its own openssl library, or something similar.

Let's try to merge and see how continuous integration react.

PR #214, opened 2020-06-14
Updated libDiscordBot description

tostc:
The libDiscordBot library is now in a useable state, so I've replaced the WIP text with the actual description of the library.


bsergean:
thanks !

Issue #224, opened 2020-07-09
Websocket destructor is not called after closing the connection


marcelkauf:
The WebSocket destructor is not called when closing a connection because the callback
`webSocket->setOnMessageCallback([webSocket, connectionState](const ix::WebSocketMessagePtr& msg){}` that is stored in `WebSocket::_onMessageCallback` is never reset. Thus we have a cyclic reference websocket -> onMessageCallback -> websocket and thus a memory leak. See #225  with the corresponding testcase.

It is not mentioned in the docs that resetting the callback is something the user has to do. And in my opinion that would not feel right since the websocket is created in the server code. I tried to reset the callback at the end of ` WebSocketServer::handleConnection` via `websocket->setOnMessageCallback(nullptr)` so that the websocket is correctly destroyed. But since this callback is also stored in `_ws.setOnCloseCallback`, this results in a segmentation fault with the workaround mentioned in #223 when a parallel worker thread tries to call the OnCloseCallback.

bsergean:
Thanks for the report and unittest @marcelkauf.

We could reset the callback in the destructor. I can't quite remember why it isn't done. One thing is that we want to invoke a callback with a 'about to destroy' or 'closing' equivalent in the destructor.

Maybe the behavior should be different for a websocket client and server object in that case. Or we would stop to manually call the close in the websocket server handle connection case.

bsergean:
Hey something to know about ; the connectionState object which is a shared pointer is passed down in a lamdbda capture, that doesn't always work well.

A good reading on it -> https://eklitzke.org/notes-on-std-shared-ptr-and-std-weak-ptr

If you have suggestions on how to fix this please bring it up.

Another thing to know: we have a set of connected clients, which might be a problem (this is helpful when implementing chat like applications to broadcast messages to all the connected clients). Maybe we could get rid of it, and leave it to the application level to handle this.

rafalsk:
What's the best workaround to handle this case? Top-notch stability an utmost priority over here.

bsergean:
I think this is fixed now @marcelkauf

marcelkauf:
Yes it is. Thanks for the fix.

Issue #223, opened 2020-07-09
Possible deadlock at connection close when running a parallel thread on the same websocket


marcelkauf:
We have the requirement to start a parallel worker thread when a connection is established to the server in order to periodically send data to the client. So the websocket is used by a thread in the ixwebsocket lib and a thread on our side. When the client closes the connection, there currently does not seem to be a proper way of ending the worker thread, at least we cannot join the thread without running into a deadlock.

The first image shows the stacktrace of the library thread when the close message of the client is received. The lib calls `setReadyState(close)`, locks `_closeDataMutex`, calls our close handler that sets a stop flag in the worker thread and joins the worker thread. But if the worker thread is currently sending data, a deadlock occurs because the send method also calls `setReadyState(close)` and tries to lock the mutex that is already locked by the websocket thread (second image).

We can solve the situation by just detaching the worker thread, but this feels more like a workaround and has other effects (see next issue). Do you have any advice for this situation?

![WebsocketThread](https://user-images.githubusercontent.com/31201841/87016753-67145c00-c1cf-11ea-8187-d5a00dd24960.png)
![WorkerThread](https://user-images.githubusercontent.com/31201841/87016771-6da2d380-c1cf-11ea-8a61-f1d7c9c4c332.png)

bsergean:
I think that this commit (22dffd5b7e5b4938a89c7182b9fc50515e95f42a) should fix this ; it eliminates the  _closeDataMutex.

I ran into a very similar dead-lock which went away after that commit.

PR #228, opened 2020-07-22
#224: Websocket destructor is not called after closing the connection

marcelkauf:
Solves #224
Use a `weak_ptr` to the websocket in `WebSocketServer::onConnectionCallback` to break the cycle `websocket -> onMessageCallback -> websocket`. This properly deletes the websocket when the connection is closed.

The handling of the `weak_ptr` is rather awkward:

```
server.setOnConnectionCallback([&webSocketPtr](std::weak_ptr<ix::WebSocket> webSocket,
                                                        std::shared_ptr<ConnectionState> connectionState,
                                                        std::unique_ptr<ConnectionInfo> connectionInfo)
{
  if (auto ws = webSocket.lock())
  {
    ws->setOnMessageCallback([webSocket, connectionState](const ix::WebSocketMessagePtr& msg)
    {
      if (msg->type == ix::WebSocketMessageType::Open)
      {
        if (auto ws = webSocket.lock())
        {
          ws->send("hello");
        }
      }
    });
  }
});
```

This solution "works for us", better alternatives would probably need a redesign of the nested `onConnectionCallback/onMessageCallback` mechanism. For example you might consider setting the `onMessageCallback` on a different entity (like the server or something else) and pass the websocket as a parameter, something like

```
server.setOnMessageCallback([](WebSocket& webSocket)
{
  if (msg->type == ix::WebSocketMessageType::Open)
  {
    webSocket.send("hello");
  }
});
```

I did not look at the unit tests (despite the leak test) and build with `cmake -DUSE_TEST=1 -DUSE_WS=1 ..`, so probably I forgot something.

Consider this pull request as work in progress and do with it whatever you want. I don't have a better solution at the moment :(.

### DIFF
--- a/ixsnake/ixsnake/IXSnakeServer.cpp
+++ b/ixsnake/ixsnake/IXSnakeServer.cpp
@@ -60,14 +60,15 @@ namespace snake
         _server.setConnectionStateFactory(factory);
 
         _server.setOnConnectionCallback(
-            [this](std::shared_ptr<ix::WebSocket> webSocket,
+            [this](std::weak_ptr<ix::WebSocket> webSocket,
                    std::shared_ptr<ix::ConnectionState> connectionState,
                    std::unique_ptr<ix::ConnectionInfo> connectionInfo) {
                 auto state = std::dynamic_pointer_cast<SnakeConnectionState>(connectionState);
                 auto remoteIp = connectionInfo->remoteIp;
-            
-                webSocket->setOnMessageCallback(
-                    [this, webSocket, state, remoteIp](const ix::WebSocketMessagePtr& msg) {
+                auto ws = webSocket.lock();
+                if (ws)
+                {
+                    ws->setOnMessageCallback([this, webSocket, state, remoteIp](const ix::WebSocketMessagePtr& msg) {
                         std::stringstream ss;
                         ix::LogLevel logLevel = ix::LogLevel::Debug;
                         if (msg->type == ix::WebSocketMessageType::Open)
@@ -115,11 +116,16 @@ namespace snake
                         else if (msg->type == ix::WebSocketMessageType::Message)
                         {
                             ss << "Received " << msg->wireSize << " bytes" << std::endl;
-                            processCobraMessage(state, webSocket, _appConfig, msg->str);
+                            auto ws = webSocket.lock();
+                            if (ws)
+                            {
+                                processCobraMessage(state, ws, _appConfig, msg->str);
+                            }
                         }
 
                         ix::CoreLogger::log(ss.str().c_str(), logLevel);
                     });
+                }
             });
 
         auto res = _server.listen();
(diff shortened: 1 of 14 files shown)


bsergean:
That's great / thanks for fixing this.

Like you said I wish I could make it less awkward for users, and I've been trying to think about a better interface, but I might merge your change in the meantime.

bsergean:
Hey I've made a change in 2798886c0bda9e1e64e1665ff97e2351a29310e7 to add a new API which is easier to use, and uses references, which was your second suggestion for improvement. For now the other API is still allowed, but if the new one works I might deprecate it or rather remove it.

I changed the usage.md file to describe on how to use it, and it is used in the echo_server.cpp sample.
https://github.com/machinezone/IXWebSocket/blob/master/ws/ws_echo_server.cpp#L45

Please tell me if it makes sense to you.

marcelkauf:
Sorry, I passed the websocket by reference only because I was too lazy to use a pointer type. The reference parameter only works after #223 has been fixed. We use the websocket in parallel by another thread and cannot join that thread when receiving a close message due to the deadlock situation described in the issue. So we have to detach that thread, but then the websocket  will probably be deleted before the thread finishes when using a reference parameter.

So as long as this problem exists, a `shared_ptr` would probably be the best type. Something like this should work (although I am getting a `bad_alloc` in debug mode when copying `msg->str` in our callback, which I didnt get before. That is really strange):

```
using OnClientMessageCallback = std::function<void(std::<shared_ptr> WebSocket, const WebSocketMessagePtr&)>;

webSocket->setOnMessageCallback(
                [this, webSocket = std::weak_ptr(webSocket)](const WebSocketMessagePtr& msg) {
                   auto ws = webSocket.lock();
                   if (ws) {
                    _onClientMessageCallback(ws, msg);
                   }
                });
```
This at least achieves both avoiding the reference cycle and keeping the socket alive if it is used somewhere else.

By the way, with the new API the websocket cannot be configured anymore (e.g. `setMessageDeflateOptions()`), because the code would be called for every incoming message.

bsergean:
Let me see if I can now address #223. I just pushed c26a2d5d39553b5ba9ac73da92be747fe9a68bc1 will that help ?

I had not thought about the configuration problem. It would be a bit annoying, but we could have some global server options to disable zlib compression (there is a problem with that btw in server mode).

bsergean:
c26a2d5d39553b5ba9ac73da92be747fe9a68bc1 is not enough I'm getting runtime crashes in the unittest on Linux / Maybe I need to reset some other callback (the other onnclose one you mentioned).

bsergean:
I think it's expected that the unittest fails, it's for the same reason as you ; the old api is still being used by the unittest.

I think I can keep the new api as it, and convert the old api to your current style with the weak_ptr.

bsergean:
So I implemented this in 847fc14 (sorry I did not try to use your pull request code so that we can get your name in the git history ...), but most other commands and unittest use the new api.

That means that you should be able to use the mainline version and not your fork anymore.

Issue #243, opened 2020-09-12
Http Server - Ability to access the payload of a POST request?


crjc:
Hi,

I've been playing around with this excellent library, am I right that it's currently not possible to access the payload of a request with the HttpServer? 

If not, is there some specific reason for this or would it be a simple feature to add?

bsergean:
The code for the http server starts here:

    void HttpServer::handleConnection(std::unique_ptr<Socket> socket,          
                                      std::shared_ptr<ConnectionState> connectionState) 
    {                                 
        _connectedClientsCount++;                                              
    
        auto ret = Http::parseRequest(socket);                                 
        // FIXME: handle errors in parseRequest                                
        
        if (std::get<0>(ret))
        {                                                                      
            auto response = _onConnectionCallback(std::get<2>(ret), connectionState);       
            if (!Http::sendResponse(response, socket))                         
            {
                logError("Cannot send response");                              
            }
        }   
        connectionState->setTerminated();
            
        _connectedClientsCount--;                                              
    }   

So the place to change would be parseRequest, whose code is in IXHttp.cpp

    std::tuple<bool, std::string, HttpRequestPtr> Http::parseRequest(
        std::unique_ptr<Socket>& socket)
    {
        HttpRequestPtr httpRequest;

        std::atomic<bool> requestInitCancellation(false);

        int timeoutSecs = 5; // FIXME

        auto isCancellationRequested =
            makeCancellationRequestWithTimeout(timeoutSecs, requestInitCancellation);

        // Read first line
        auto lineResult = socket->readLine(isCancellationRequested);
        auto lineValid = lineResult.first;
        auto line = lineResult.second;

        if (!lineValid)
        {
            return std::make_tuple(false, "Error reading HTTP request line", httpRequest);
        }

        // Parse request line (GET /foo HTTP/1.1\r\n)
        auto requestLine = Http::parseRequestLine(line);
        auto method = std::get<0>(requestLine);
        auto uri = std::get<1>(requestLine);
        auto httpVersion = std::get<2>(requestLine);

        // Retrieve and validate HTTP headers
        auto result = parseHttpHeaders(socket, isCancellationRequested);
        auto headersValid = result.first;
        auto headers = result.second;

        if (!headersValid)
        {
            return std::make_tuple(false, "Error parsing HTTP headers", httpRequest);
        }

        httpRequest = std::make_shared<HttpRequest>(uri, method, httpVersion, headers);
[...]

PR #244, opened 2020-09-12
Read body in parseRequest for HttpServer

carr-7:
PR prompted from my question #243 

Simply reads the request's body into a std::string if the Content-Length header was set.


bsergean:
Thanks !

I will reformat slightly the code (I use a different style) and use a readBytes (plural) which should be more efficient.

PR #245, opened 2020-09-12
Rename HttpResponse's payload to body

carr-7:
I think it feels odd/inconsistent using the name `payload` in `ix::HttpResponse`, especially when the name `body` is used in `ix::HttpRequest`. 

Would renaming this be OK?


bsergean:
That's fine, it's consistent now which is better.

Can you try to fix the ws command line tool, ws.cpp in the ws folder ? It still has some references to payload.

```
ws$ ag payload
ws.cpp
1457:                out.write((char*) &response->payload.front(), response->payloa.size());
1464:                    spdlog::info("payload: {}", response->payload);
2549:                        spdlog::info("payload: {}", response->payload);
2557:                    spdlog::error("Response: {}", response->payload);
```

bsergean:
Oh and under the docs folder in the usage there is likely references to payload. Can you change them to body ?

bsergean:
Thanks !

If you can say how you are using the library feel free to edit the README.md (see at the bottom) or I can do that for you if you tell me.

PR #247, opened 2020-10-19
Fix an issue with disabling zlib and getting linker errors from the http client.

SeanOMik:
I was running into linker errors when I would disable zlib and try to use the HTTP client. After I looked at them, it looked like it was trying to use zlib to compress requests (only when args->compress was set), this caused linker errors.

I fixed them by adding `#ifdef IXWEBSOCKET_USE_ZLIB` before the if statements and also where you define the `compress` and `compressRequest` fields in the `ix::HttpRequestArgs` struct since they won't be used without zlib.

I tried to follow the same commit message format that I saw you use, in this commit.

### DIFF
--- a/ixwebsocket/IXHttp.h
+++ b/ixwebsocket/IXHttp.h
@@ -83,8 +83,10 @@ namespace ix
         bool followRedirects = true;
         int maxRedirects = 5;
         bool verbose = false;
+#ifdef IXWEBSOCKET_USE_ZLIB
         bool compress = true;
         bool compressRequest = false;
+#endif
         Logger logger;
         OnProgressCallback onProgressCallback;
     };
--- a/ixwebsocket/IXHttp.h
+++ b/ixwebsocket/IXHttp.h
@@ -83,10 +83,8 @@ namespace ix
         bool followRedirects = true;
         int maxRedirects = 5;
         bool verbose = false;
-#ifdef IXWEBSOCKET_USE_ZLIB
         bool compress = true;
         bool compressRequest = false;
-#endif
         Logger logger;
         OnProgressCallback onProgressCallback;
     };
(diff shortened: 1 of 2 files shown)


ixwebsocket/IXHttp.h
```diff
@@ -83,8 +83,10 @@ namespace ix
         bool followRedirects = true;
         int maxRedirects = 5;
         bool verbose = false;
```
bsergean, line 85: I know it kindof makes sense, but I prefer if we keep that one block as is.

ixwebsocket/IXHttpClient.cpp
```diff
@@ -6,7 +6,9 @@
 
 #include "IXHttpClient.h"
```
bsergean, line 8: Same here / we should be able to include this header.

```
#pragma once

#include <string>

namespace ix
{
    std::string gzipCompress(const std::string& str);
    bool gzipDecompress(const std::string& in, std::string& out);
} // namespace ix
```

ixwebsocket/IXHttpClient.cpp
```diff
@@ -204,11 +206,13 @@ namespace ix
         if (verb == kPost || verb == kPut || verb == kPatch || _forceBody)
         {
             // Set request compression header
```
bsergean, line 208: This is good.

ixwebsocket/IXHttpClient.cpp
```diff
@@ -580,10 +584,12 @@ namespace ix
                 multipartBoundary, httpFormDataParameters, httpParameters);
         }
```
bsergean, line 586: This is good too.

bsergean:
To make a change

1. type make change (or edit the VersionFile and the Changelog file)
2. type make commit

SeanOMik:
Okay, I made those changes

bsergean (approved):
All good

bsergean:
Thanks for the fix. I tagged a new version (v10.5.3) and added a note in the changelog.

PR #249, opened 2020-11-01
Add project to users list

luccanunes:


bsergean:
Thanks @luccanunes !

bsergean:
It's funny lots of people make discord bots with this code. Please report bugs if there library prevent you from doing it, but given that a lot of folks did it ... it must be good enough :)

luccanunes:
It certainly is great. 

I think the main reason this is awesome for making discord libraries, is that it's very simple to use, and, as the library maker, you can't worry that much about all the networking and still manage all the library structure.
I haven't encountered any bugs so far, it seems really stable. I'll let you guys know if I ever do.

Thanks for making such a nice library and also for the merge!

bsergean:
Thanks for the good words ;)

PR #250, opened 2020-11-10
Always make OpenSSL verify the peer certificate

jb-gcx:
**Error description:**

When using an `IXWebSocket` with the default SSL config, the server certificate is not checked. This can be reproduced by setting your system clock to a date that is outside of the certificates begin/end date. The `IXWebSocket` (specifically `IXSocketOpenSSL`) will still happily connect to it. The OpenSSL version used is 0x1010107fL (1.1.1g).

**Fix:**

Always set `SSL_VERIFY_PEER`, unless `_tlsOptions.isPeerVerifyDisabled()` returns `true`.

**Caveat:**

I don't know if this routine is also used for server sockets. In that case there needs to be a special case, because the `IXWebSocketServer` validating a client certificate is probably not desired. Is this a problem?

### DIFF
--- a/ixwebsocket/IXSocketOpenSSL.cpp
+++ b/ixwebsocket/IXSocketOpenSSL.cpp
@@ -503,14 +503,14 @@ namespace ix
                         errMsg += ERR_error_string(sslErr, nullptr);
                         return false;
                     }
-
-                    SSL_CTX_set_verify(
-                        _ssl_context, SSL_VERIFY_PEER, [](int preverify, X509_STORE_CTX*) -> int {
-                            return preverify;
-                        });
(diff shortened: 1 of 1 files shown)


bsergean (approved):
This looks like a regression from 677f79b0ea87ececad2bb9295630f4b140ec415b, and the server code which does that in the accept method isn't affected.

CryptoManiac (commented):
This seems to be breaking compatibility with at least four SSL servers I have actually seen in production.

Examples are these:

wss://btc1.trezor.io
wss://btc2.trezor.io
wss://btc3.trezor.io

Simply try to connect and you will get this message:

```OpenSSL failed error:16000069:STORE routines::unregistered scheme```

PR #264, opened 2021-01-11
Allow customizing the websocket handshake timeout

mrexodia:


ixwebsocket/IXWebSocket.h
```diff
@@ -48,6 +48,7 @@ namespace ix
         void setUrl(const std::string& url);
 
         // send extra headers in client handshake request
```
bsergean, line 50: Could you move the comment 'send extra headers ...' next to the setExtraHeader method ?
And search the doc to see if we can document the new method ?

PR #266, opened 2021-01-17
add a option that enables user to set any default reconnection interval

masakiishii:
I added a option that enables user to set any default reconnection interval. In the current version, IXWebsocket tries to reconnect immediately if the connection close unexpectedly(I mean, the first attempt reconnection event before calculating the wait time by exponential backoff algorithm), but depending on applications, I think, on that same situation, some users want to set any intervals for the first attempt and any constant additional time to the wait time which the exponential backoff algorithm calculate, so that's why I sent pull request.

### DIFF
--- a/ixwebsocket/IXWebSocket.cpp
+++ b/ixwebsocket/IXWebSocket.cpp
@@ -22,6 +22,7 @@ namespace ix
     const int WebSocket::kDefaultPingIntervalSecs(-1);
     const bool WebSocket::kDefaultEnablePong(true);
     const uint32_t WebSocket::kDefaultMaxWaitBetweenReconnectionRetries(10 * 1000); // 10s
+    const uint32_t WebSocket::kDefaultReconnectionInterval(0);
 
     WebSocket::WebSocket()
         : _onMessageCallback(OnMessageCallback())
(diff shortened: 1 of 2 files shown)


bsergean:
Hi @masakiishii thanks for the contribution.

I'm not sure this will work as is, and it would be better to have documentation on the feature.
We would probably want an enum to describe / get / set all the different reconnections technique.

1. backoff xxx (current default one)
2. periodical interval one (the one you added)

setReconnectionAlgorithm()
getReconnectionAlgorithm()

masakiishii:
Hi @bsergean, thank you for your reply.

I agree with your proposal. I also think it would be better to enable users to combine these enum.

**Pattern1. exponential backoff (as is)**

        1. wait time = 0                 (where first reconnection, _**not first connection**_.)

        2. wait time = (2^retries * 100) (where retries > 0, this is exponential backoff algorithm)


(*) This "retries" means "reconnection retries", used so in IXWebsocket.

The constant term 'a' in following formula is expressing periodical invervals.

**Pattern2. periodical interval**

        1. wait time = a (where **_not first connection_**.)

**Pattern3. Pattern1 + Pattern2**

        1. wait time = a                       (where first reconnection, **_not first connection_**.)

        2. wait time = (2 ^ retries * 100) + a (where retries > 0)

The pattern of the pull request I send before is 3.

bsergean:
I just fixed a bug in ws connect.

Can you explain what was your main problem ? Were the early reconnect happening too early ? (800ms, 400ms, etc...). Maybe we could have a 'first reconnect time ... that isn't too well defined, but that would be the first value which is then used for the exponential backoff.

bsergean:
I just added a new option -> d26664f which actually should be just like your **Pattern 1 + Pattern 2**.

Can you try to use setMinWaitBetweenReconnectionRetries ?

masakiishii:
Thank you for your modification. I tried and the behavior is as I expected.

bsergean:
Thanks for testing @masakiishii / I'm gonna close this as it looks like you have everything you need now, if you find something else please go ahead and make a ticket or a PR. Have a good day !

PR #278, opened 2021-03-19
Adds demo building option.

innerNULL:
Adds a demo-building option in CMakeLists.txt to make more convenient.


bsergean:
Thanks @innerNULL / I might tweak that PR fyi.

PR #279, opened 2021-03-26
Fixed example code for the new API of IXWebSocketServer

brt-v:
The syntax of the WebSocketServer example was invalid. Not sure if it's perfect, but at least it compiles and runs.


bsergean:
Thanks for the contribution !

PR #286, opened 2021-04-16
(http server) Do not overwrite response headers with request headers

Anna-Boettcher-JPL:
When trying to open files served on a local machine with Firefox 87 on Ubuntu 18.04, I discovered a bug where the HttpResponse was overwriting all of its headers with anything in HttpRequest.

In particular this browser had a header `Content-Encoding:` in the request which was eliminating the gzip option specified by the server, but it also doesn't make sense in general to overwrite the response headers with the request headers that have been explicitly set.

### DIFF
--- a/ixwebsocket/IXHttpServer.cpp
+++ b/ixwebsocket/IXHttpServer.cpp
@@ -151,7 +151,11 @@ namespace ix
 
                 for (auto&& it : request->headers)
                 {
-                    headers[it.first] = it.second;
+                    //Do not replace our explicitly set server headers with client headers
+                    if(!headers.count(it.first))
+                    {
+                        headers[it.first] = it.second;
+                    }
                 }
 
                 return std::make_shared<HttpResponse>(


bsergean:
Thanks for the PR. Could you add a comment in code that describe why we're doing this.

It looks like the server shouldn't win, and trust the client http headers ? Can you check if there's the same problem in the websocket code path ?

Anna-Boettcher-JPL:
The server shouldn't win? This is the exact opposite - the server is allowing the client to dictate its headers over its own. Which makes no sense. In this case it breaks the server.

Anna-Boettcher-JPL:
as far as I can tell this is not the case for WS, only HTTP server

bsergean:
1. Can you add a new test-case in test/IXHttpServerTest.cpp, or update an existing one to cover this ?
2. Can you update formatting ? If you have clang-format installed you can type `make format`, or simply look around in the code base. spaces are required after if (something), and after comments // something.

Anna-Boettcher-JPL:
Sorry, I really am just contributing a fix to an issue. I will simply maintain a fork with this fix since I don't have time for this anymore.

bsergean:
No worries, thanks for the contribution @Anna-Boettcher-JPL / I'll try to fix that in the future.

glenne:
@bsergean Why is this closed?  It's still a bug today.  WIthout the fix, all files delivered when USE_GZIP=1  by the httpserver to a browser do not have the Content-Encoding=gzip set as it is overwritten by the client request.  This causes the browser to display raw gzip gobbletgook instead of rendering html.  Please merge the pull request.

bsergean:
See my previous comment @glenne 

    1. Can you add a new test-case in test/IXHttpServerTest.cpp, or update an existing one to cover this ?
    2. Can you update formatting ? If you have clang-format installed you can type make format, or simply look around in the code base. spaces are required after if (something), and after comments // something.

PR #294, opened 2021-06-03
Fix unsafe calls and safeguard WebSocketMessage

picanumber:
Binding `WebSocketMessage` to a temporary message argument can create undefined behavior. This MR introduces a deleted constructor overload to "absorb" cases where string arguments are temporaries and fixes unsafe calls that were "exposed" by the deleted constructor.

### DIFF
--- a/ixwebsocket/IXWebSocket.cpp
+++ b/ixwebsocket/IXWebSocket.cpp
@@ -15,7 +15,10 @@
 #include <cmath>
 
 
-static const std::string emptyMsg;
+namespace
+{
+    const std::string emptyMsg;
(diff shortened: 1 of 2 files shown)


ixwebsocket/IXWebSocket.cpp
```diff
@@ -15,6 +15,9 @@
 #include <cmath>
```
bsergean, line 17: Can you put that emptyMsg constants in the anonymous namespace ?

bsergean:
Thanks @picanumber / merged and released v11.2.8

PR #296, opened 2021-06-04
Avoid duplicate Accept and User-Agent headers

crjc:
The logic for preventing these duplicate headers was flawed. It was checking the empty response's headers rather than the headers specified in the request.


bsergean:
thanks @crjc !

Issue #292, opened 2021-05-31
Unsafe locking pattern


picanumber:
There is faulty locking logic at least in a couple spots in `IXWebSockets.cpp`.  Take for example the following function:

```cpp
const std::string& WebSocket::getUrl() const
{
    std::lock_guard<std::mutex> lock(_configMutex);
    return _url;
}
```

There is no point in exclusive access when returning a reference:

1. The caller will get a seemingly `const` object but that object can be modified by a different thread, even worse be in an intermediate state while a different thread is modifying it.
2. If there is no chance a different is modifying the object, you don't need to protect it with a lock. In any case the lock only protects `_url` while inside the member function, but any protection goes away after the caller gets the reference. Bottom line is that the lock inside that function does little to nothing.

Since there are reader/writer methods it would make sense to use a `shared_mutex` + `unique_lock` to get less aggressive locking as a first step and then return values instead of references in the functions that retrieve internal state. 

Would a Pull request with such a refactoring be of interest?

bsergean:
Thanks for the report. The lock was put in place to silence thread sanitizer / tsan.

If you want to improve that code I'd be happy to look at your PR, the constraint is to keep the code C++11 compatible. Returning values sounds good.

PR #297, opened 2021-06-04
Avoid returning references that are mutex protected

picanumber:
Motivation for this MR #292 
 
 The antipattern of returning references to mutex protected members was removed. Since a caller can hold the reference it would make all class level locking meaningless.
 
 Instead values are returned. The IXWebSocketPerMessageDeflateOptions class was shrunk by 7 bytes (1 padding + 2*3) after changing the int members to the used uint8_t; side effects of that were handled.
 
 An inefficient "string -> int" was replaced by standard library. As seen here http://coliru.stacked-crooked.com/a/46b5990bafb9c626 this gives an order of magnitude better performance.

### DIFF
--- a/ixwebsocket/IXWebSocket.cpp
+++ b/ixwebsocket/IXWebSocket.cpp
@@ -41,7 +41,8 @@ namespace ix
         , _pingIntervalSecs(kDefaultPingIntervalSecs)
     {
         _ws.setOnCloseCallback(
-            [this](uint16_t code, const std::string& reason, size_t wireSize, bool remote) {
+            [this](uint16_t code, const std::string& reason, size_t wireSize, bool remote)
+            {
                 _onMessageCallback(
                     ix::make_unique<WebSocketMessage>(WebSocketMessageType::Close,
                                                       emptyMsg,
@@ -75,7 +76,7 @@ namespace ix
         _extraHeaders = headers;
     }
 
-    const std::string& WebSocket::getUrl() const
+    const std::string WebSocket::getUrl() const
     {
         std::lock_guard<std::mutex> lock(_configMutex);
         return _url;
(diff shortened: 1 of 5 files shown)


ixwebsocket/IXWebSocket.cpp
```diff
@@ -41,7 +41,8 @@ namespace ix
         , _pingIntervalSecs(kDefaultPingIntervalSecs)
     {
         _ws.setOnCloseCallback(
-            [this](uint16_t code, const std::string& reason, size_t wireSize, bool remote) {
```
picanumber, line 44: This minor formatting change stems from simply applying clang format with the projects `.clang-format` configuration.

ixwebsocket/IXWebSocketPerMessageDeflateOptions.cpp
```diff
@@ -85,11 +85,7 @@ namespace ix
 
             if (startsWith(token, "server_max_window_bits="))
             {
```
bsergean, line 87: stoi returns 0 on errors right ?

bsergean (approved):

bsergean:
thanks !

ixwebsocket/IXWebSocketPerMessageDeflateOptions.cpp
picanumber: No, it throws an exception (invalid arg for non number strings, or out of range for non convertible argument). Even if you handle exceptions locally and you have a "high" exceptional case ratio, you still remain an order of magnitude faster than the `stringstream` solution http://coliru.stacked-crooked.com/a/abc62fdec42443f5

I'll make a quick MR in case you want to review how such local handling would look like (I mean in case you don't handle exceptions further down the execution stack).

PR #299, opened 2021-06-05
Noexcept ix web socket per message deflate options

picanumber:
Since there's no explicit exception handling, this MR suggests a `noexcept` alternative to `std::stoi` (basically using its internal implementation directly)


bsergean:
Thanks, I think it's better that way.

PR #307, opened 2021-07-14
Only find OpenSSL, MbedTLS, zlib if they have not already been found, make CMake install optional.

mattparks:


bsergean:
Thanks for the fix. I'd like to merge it, but first I would want to be sure it does not break the github actions. Are they still passing in your fork ?

bsergean:
@mattparks could rebase your branch of master.

I've enabled github actions so that they will run for your change.
https://github.com/machinezone/IXWebSocket/pull/313

Some actions are flaky, but if most of them pass I'll merge your pr.

PR #318, opened 2021-09-20
Use GNUInstallDirs in cmake

### LINKED ISSUE #317: [Linux] CMake: Use GNUInstallDirs instead of hardcoded libdir/includedir
I just happened to try installing IXWebSocket under a Linux system with `/usr/lib64` as libdir. However, the CMakeLists.txt has `${CMAKE_INSTALL_PREFIX}/lib` hardcoded as libdir/includedir, causing the library and CMake module to get installed in the wrong directory:

https://github.com/machinezone/IXWebSocket/blob/dbafa0aa07255d50b00c91c2f8b8a2825647447c/CMakeLists.txt#L245-L248

https://github.com/machinezone/IXWebSocket/blob/dbafa0aa07255d50b00c91c2f8b8a2825647447c/CMakeLists.txt#L252-L256

https://github.com/machinezone/IXWebSocket/blob/dbafa0aa07255d50b00c91c2f8b8a2825647447c/CMakeLists.txt#L258-L261

I don't know if these are all occurences of a hard-coded libdir/includedir, but my attempt at installing the library after patching those lines did the job.

I recommand using [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) instead of hardcoding `include` and `lib` as includedir/libdir. This way, you don't have to bother about determinig the correct includedir/libdir yourself. Just `include(GNUInstallDirs)` and replace `include` with `${CMAKE_INSTALL_INCLUDEDIR}` and `lib` with `${CMAKE_INSTALL_LIBDIR}`.

NexAdn:
This PR uses GNUInstallDirs to determine the correct installation includedir/libdir for libraries and headers in CMake.

Tested successfully unter Gentoo Linux, but should work on all other Linux distributions as well, since it's just a standard installation procedure used by many other projects.

Fixes #317.

Signed-off-by: NexAdn <git@nexadn.de>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,23 +242,25 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
     target_compile_options(ixwebsocket PRIVATE /MP)
 endif()
 
+include(GNUInstallDirs)
+
 target_include_directories(ixwebsocket PUBLIC
   $<BUILD_INTERFACE:${IXWEBSOCKET_INCLUDE_DIRS}/>
-  $<INSTALL_INTERFACE:include/ixwebsocket>
+  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/ixwebsocket>
 )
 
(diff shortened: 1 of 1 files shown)


bsergean:
Thanks !

PR #325, opened 2021-11-15
fix compilation under mingw64

### LINKED ISSUE #324: compilation failure under MINGW64
IXWebSocket seems to work perfectly fine under [MSYS2/MINGW64](https://www.msys2.org/) but theres a couple changes i have to make to get it to compile

IXNetSystem.cpp fails due to `snprintf` and `uint16_t` not being found which i fixed by including `cstdio` and `cstdint`

IXNetSystem.h fails due to the [redefinition](https://github.com/machinezone/IXWebSocket/blob/3f1fc6906c64ee97a5b7e5b015ef29120af3481b/ixwebsocket/IXNetSystem.h#L37-L51) of `pollfd`

it seems the github actions set up uses version 8.1.0 of mingw-w64 off choco where the definition is necessary but newer versions seem to define pollfd so it fails to compile. not really sure what the best idea to go about this would be but if you wanna allow 8.1.0 to keep working perhaps you can check the GCC version macros, or i would imagine CMake provides a method to determining if pollfd exists so that it can be defined if it hasnt been already (maybe [this?](https://cmake.org/cmake/help/latest/module/CheckTypeSize.html#module:CheckTypeSize))

ouwou:
closes #324

i dug a little deeper and found [why](https://github.com/mingw-w64/mingw-w64/blob/564d8771316bc1836c5e35dae16a5f87467eb310/mingw-w64-headers/include/winsock2.h#L1116) pollfd wasnt being defined, so hopefully this should take care of all situations

### DIFF
--- a/ixwebsocket/IXNetSystem.cpp
+++ b/ixwebsocket/IXNetSystem.cpp
@@ -5,6 +5,8 @@
  */
 
 #include "IXNetSystem.h"
+#include <cstdint>
+#include <cstdio>
 
 namespace ix
 {
--- a/ixwebsocket/IXNetSystem.h
+++ b/ixwebsocket/IXNetSystem.h
@@ -34,8 +34,8 @@
 // Define our own poll on Windows, as a wrapper on top of select
 typedef unsigned long int nfds_t;
 
-// mingw does not know about poll so mock it
-#if defined(__GNUC__)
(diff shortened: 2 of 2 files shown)


bsergean:
This is looking good, thanks for the fix !

PR #327, opened 2021-11-23
Add getters

### LINKED ISSUE #326: port, host... in IXWebSocket.hare private with no getters
Hello,
Thanks for this nice library.

I try to access the host, port etc of WebSocketServer but there is no getters for them. Even child class cannot access to them because they are private on `IXSocketServer.h`

Would you accept some PR with these getters ?

flagarde:
Fix #326

### DIFF
--- a/ixwebsocket/IXSocketServer.h
+++ b/ixwebsocket/IXSocketServer.h
@@ -60,6 +60,11 @@ namespace ix
 
         void setTLSOptions(const SocketTLSOptions& socketTLSOptions);
 
+        int  getPort();
+        std::string getHost();
+        int getBacklog();
+        std::size_t getMaxConnections();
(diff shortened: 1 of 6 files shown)


ixwebsocket/IXSocketServer.h
```diff
@@ -60,6 +60,11 @@ namespace ix
 
         void setTLSOptions(const SocketTLSOptions& socketTLSOptions);
 
+        int  getPort();
+        std::string getHost();
+        int getBacklog();
+        std::size_t getMaxConnections();
```
bsergean, line 66: std::size_t is a bit verbose but maybe it is always defined on Windows. I had some weird macro for it in other places in the code base.

bsergean:
Thanks !

PR #332, opened 2021-12-08
Allow to cancel asynchronous HTTP requests

natano:
Usage:

```cpp
	auto args = this->httpClient.createRequest(url, method);
	httpClient.performRequest(args, ...);
	[...]
	// Oops, we don't actually want to complete the request!
	args->cancel = true;
```

### DIFF
--- a/ixwebsocket/IXHttpClient.cpp
+++ b/ixwebsocket/IXHttpClient.cpp
@@ -241,17 +241,21 @@ namespace ix
         std::string errMsg;
 
         // Make a cancellation object dealing with connection timeout
-        auto isCancellationRequested =
-            makeCancellationRequestWithTimeout(args->connectTimeout, _stop);
+        auto cancelled = makeCancellationRequestWithTimeout(args->connectTimeout, args->cancel);
+
+        auto isCancellationRequested = [&]() {
+            return cancelled() || _stop;
+        };
 
         bool success = _socket->connect(host, port, errMsg, isCancellationRequested);
         if (!success)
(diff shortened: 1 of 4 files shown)


ixwebsocket/IXHttpClient.cpp
```diff
@@ -260,7 +263,7 @@ namespace ix
         }
```
bsergean, line 264: For the transfer timeout, _stop will not be read which sounds incorrect.

Also can you edit the doc to add description for your new cancel argument ?

bsergean: I think we just need to replicate the logic you added with your lambda down here.

Also I think it would be good to have a unittest for that new code.

natano: I added a unit test + a special error (`HttpErrorCode::Cancelled`) to indicate that the request was cancelled and I also updated the docs to show how cancellation is used.

According to my understanding of C++ semantics it shouldn't be necessary to replicate the lambda further down. It has a reference to the `cancelled` variable which we update later. When `cancelled` is reassigned that change is reflected in the lambda (because it's a reference and not a copy).

bsergean: Thanks, I'll take a look.

bsergean (commented):
Looking good, thanks for adding the unittest.

bsergean:
The unittest passed on all platform, merging. Thanks for the contribution !

PR #336, opened 2021-12-17
Fix errors in example code.

matthewscottgordon:
Update to documentation. I tried to use this example code and it wouldn't compile; looks like there's a mistake in the name of a field.


bsergean (approved):
Thanks !

PR #339, opened 2021-12-26
Export symbols into .def files for MSVC

### LINKED ISSUE #335: building dynamic libraries on the Windows platform?
I tried to set BUILD_SHARED_LIBS to open, but the compiled library does not export functions, what should I do? thank you.

CryptoManiac:
Fix #335

https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,10 +148,29 @@ if (USE_TLS)
     endif()
 endif()
 
-add_library( ixwebsocket
-    ${IXWEBSOCKET_SOURCES}
-    ${IXWEBSOCKET_HEADERS}
-)
+if(BUILD_SHARED_LIBS)
+    # Building shared library
+    
+    if(MSVC)
+        # Workaround for some projects
+        set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -148,10 +148,29 @@ if (USE_TLS)
...
+    if(MSVC)
+        # Workaround for some projects
+        set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+    endif()
+
+    add_library( ixwebsocket SHARED
+        ${IXWEBSOCKET_SOURCES}
+        ${IXWEBSOCKET_HEADERS}
+    )
+
```
bsergean, line 163: That part will be annoying, we could try to read it from the ixwebsocket/IXVersion.h (forgot the name) instead.

CryptoManiac: It’s better then to move version definition to cmake file and then use female constants in the source files to avoid hard-coded definitions. It’s impractical to parse include file when it could be done in opposite way.

bsergean:
Thanks !

PR #342, opened 2021-12-30
Win wsa select event

ahausladen:
This PR implements a poll() function that uses WSAEventSelect, WSAWaitForMultipleEvents and WSAEnumNetworkEvents instead of select() under Windows. With the additional SelectInterruptEvent class the Socket::wakeUpFromPoll() method can now wake up the background thread and signal the "SendRequest" and "CloseRequest". Neither the first nor the last thing happend under Windows. Wake up was not possible and requests weren't processed because there was no wake up pipe with the request-value.

Details:
The ix::poll() function now takes an additional `void** event` parameter that is set to `nullptr` when the function returns if the event wasn't signaled. For non-Windows platforms the `event` parameter is always `nullptr`. If no "event" is specified the old `select` implementation is used (There is a hard coded `ix::make_unique<SelectInterrupt>()` in SocketConnect::connectToAddress that doesn't support the file descriptor nor the new event, but doesn't need it anyway)

The new SelectInterruptEvent class can also be used in non-Windows and non-Unix platforms, but then it falls back to a poll-timeout (20ms) implementation (that was the first thing I wrote to fix the issue, so I kept it in the code).

### DIFF
--- a/ixwebsocket/IXNetSystem.cpp
+++ b/ixwebsocket/IXNetSystem.cpp
@@ -183,7 +183,12 @@ namespace ix
                             break;
                         }
                     }
-                    count++;
+
+                    if (fd->revents != 0)
+                    {
+                        // only signaled sockets count
+                        count++;
+                    }
                 }
                 socketIndex++;
             }
--- a/ixwebsocket/IXNetSystem.cpp
+++ b/ixwebsocket/IXNetSystem.cpp
@@ -7,6 +7,9 @@
 #include "IXNetSystem.h"
 #include <cstdint>
 #include <cstdio>
+#ifdef _WIN32
+#include <vector>
+#endif
 
 namespace ix
 {
@@ -37,6 +40,44 @@ namespace ix
 #endif
     }
 
+#ifdef _WIN32
+    struct WSAEvent
+    {
+    public:
+        WSAEvent(struct pollfd* fd)
+            : _fd(fd)
+        {
+            _event = WSACreateEvent();
+        }
+
+        WSAEvent(WSAEvent&& source) noexcept
+        {
+            _event = source._event;
+            source._event = WSA_INVALID_EVENT; // invalidate the event in the source
+            _fd = source._fd;
+        }
+
+        ~WSAEvent()
+        {
+            if (_event != WSA_INVALID_EVENT) WSACloseEvent(_event);
+        }
+
+        operator HANDLE()
+        {
+            return _event;
+        }
+
+        operator struct pollfd*()
+        {
+            return _fd;
+        }
+
+    private:
+        HANDLE _event;
+        struct pollfd* _fd;
+    };
+#endif
+
     //
     // That function could 'return WSAPoll(pfd, nfds, timeout);'
     // but WSAPoll is said to have weird behaviors on the internet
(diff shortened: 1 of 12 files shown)


bsergean:
Great, I'm glad we can finally fix this properly on Windows. I'll have a look when I have a chance.

ixwebsocket/IXSelectInterruptEvent.cpp
```diff
@@ -0,0 +1,82 @@
+/*
+ *  IXSelectInterruptEvent.cpp
+ */
+
+//
+// On Windows we use Events to wake up select. And on any other platform
+// other than Unix we emulate by using a timeout (see )
```
bsergean, line 7: see ... comment is incomplete.

ixwebsocket/IXSelectInterruptFactory.cpp
```diff
@@ -20,7 +20,7 @@ namespace ix
 #if defined(__linux__) || defined(__APPLE__)
```
bsergean, line 20: I think we should reverse those ifdefs so that this code compiles on other unix (say freebsd etc...).

```
#ifdef _WIN32
#include "IXSelectInterruptEvent.h"
#else
#include "IXSelectInterruptPipe.h"
...
```

ixwebsocket/IXSelectInterruptFactory.cpp
```diff
@@ -20,7 +20,7 @@ namespace ix
...
         return ix::make_unique<SelectInterruptPipe>();
 #else
```
bsergean, line 22: same here.

bsergean (approved):
Very nice. I added a small comment.

ixwebsocket/IXSelectInterruptFactory.cpp
ahausladen: The SelectInterruptEvent code works for all platforms because it has a fall back "timeout based" branch if _WIN32 isn't defined. That's what the "(see...)" comment was for that I stopped writing in the middle of the comment.

bsergean:
```
 6/15 Test  #6: IXHttpClientTest .............................***Failed    0.35 sec
      Start  7: IXUnityBuildsTest
```

Something is up with this unittest, it appears broken now. Not sure if you can follow the CI weblinks.

ahausladen:
I'll have to look into it.

ahausladen:
The HttpClient doesn't use the SelectInterruptFactory. Instead it uses a hard coded "SelectInterrupt" instance that doesn't implement either `getFd()` nor the new `getEvent()` methods.

Because of this the "timeout" fallback code for SelectInterrupt ran and due to a bug (from my PR) converted a PollResultType::Timeout into a ReadyToRead.

bsergean:
Cool / with your fixes CI pass on unix, but not on windows now.

I forgot about this ... but is there a way to have the http-client code use a select-interrupt factory like everyone else ?

ahausladen:
> is there a way to have the http-client code use a select-interrupt factory like everyone else ?

There ist.

I found the problem. The Socket-Event that is used for WSAEventSelect must be deselected from the socket or it has to stay alive while the socket exists. The later would require all code that has an "int _sockFd" to be changed to hold a pair<int, WSAEVENT>. That would be too many code changes and every developer must be aware of it.
The first solution (deselecting the socket from the event) requires not that many code changes.

ixwebsocket/IXNetSystem.cpp
```diff
@@ -59,7 +59,14 @@ namespace ix
```
bsergean, line 59: Great, the change is local then which makes this easier to maintain like you pointed out.

PR #346, opened 2022-01-06
Fixed missing header for gcc (9.3.0) compilation

ahausladen:
This PR adds the missing `#include <algorithm>` for IXSelectInterruptEvent.cpp file when compiling under gcc (9.3.0).

gcc (9.3.0) fails to compile with an unknown function std::find() because it was missing the `#include <algorithm>` that seams to be included with other headers on other compiler platforms.


bsergean:
Thanks for the fix @ahausladen !

PR #347, opened 2022-01-06
Introduction of IXWebSocketSendData

ahausladen:
This PR introduces a IXWebSocketSendData class that makes it possible to not only send std::string but also std::vector<char/uint8_t> and char* without copying them to a std::string first. With this the memory usage can be reduced if the data isn't already a std::string.

It also adds a sendUtf8Text() method that doesn't check for invalid UTF-8 characters. The caller must guarantee that the string only contains valid UTF-8 characters. This helps to reduce CPU usage for large strings where the caller already knows that it is valid UTF-8.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,7 @@ set( IXWEBSOCKET_HEADERS
     ixwebsocket/IXWebSocketPerMessageDeflateCodec.h
     ixwebsocket/IXWebSocketPerMessageDeflateOptions.h
     ixwebsocket/IXWebSocketProxyServer.h
+    ixwebsocket/IXWebSocketSendData.h
     ixwebsocket/IXWebSocketSendInfo.h
     ixwebsocket/IXWebSocketServer.h
     ixwebsocket/IXWebSocketTransport.h
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -90,6 +90,18 @@ auto result =
     });
 ```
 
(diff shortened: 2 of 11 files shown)


bsergean:
Nice feature. Can you mention it in the doc too (usage.md in the docs folder) so that users know about the new APIs ?

bsergean:
Thanks for updating the doc. I'm gonna be annoying but do you mind adding a few unittest or modify an existing one to cover the new APIs ?

An interesting one would be to send invalid utf in text mode, and check that on the receiving side the message is discarded, we don't have coverage for that but a well behaved websocket library should behave like that. This should be working as this library passed the autobahn test, but since there's no automation around this maybe it did regress (like with the problem you discovered with zlib compression).

bsergean:
Also for the doc, we should change the main page README.md (and maybe index.md + main.cpp) to 'advertise' about the new feature (maybe with just a comment, or if you have another idea I'm open to suggestions).

```
    // Read text from the console and send messages in text mode.
    // Exit with Ctrl-D on Unix or Ctrl-Z on Windows.
    while (std::getline(std::cin, text))
    {
        // we can send binary data as well as text, and use multiple types (char*, std::vector<char>, etc...)
        webSocket.send(text);
        std::cout << "> " << std::flush;
    }
```

Last idea, maybe the http class could be improved to use your new buffer object, for which a new name might be required then ... IXSendData ? / not sure if this is a good idea. This could definitely be a different PR, but if we introduce new types we might as well try to only do this once, so that if we change our mind we don't have to deal with backward compatibility.

bsergean:
Ok I think it's fine as is, thanks for contributing this @ahausladen

PR #348, opened 2022-01-13
Add an alias for ixwebsocket

flagarde:
Hello, many thanks for your great library

I'm using it with FetchContent and it's working fine but I can only refer to your library with ixwebsocket while on an install based configuration (using `find_package`) the library is accessible with `ixwebsocket::ixwebsocket`.

This PR allows to use `ixwebsocket::ixwebsocket` inn both cases


bsergean:
Looking good, thanks for the PR !

PR #349, opened 2022-01-13
Use Threads::Threads target

flagarde:
Uses `Threads::Threads` `IMPORTED` target and the recommendations from CMake https://cmake.org/cmake/help/latest/module/FindThreads.html.

It's then possible to use 'PUBLIC' 'PRIVATE' ... to `target_link_libraries` commands.


bsergean:
I'll trust you on that one.

PR #353, opened 2022-01-30
Add support for streaming transfers

natano:
This change adds onChunkCallback to the request. If defined it will be called repeatedly with the incoming data. This allows to process data on the go or write it to disk instead of accumulating the data in memory.

### DIFF
--- a/ixwebsocket/IXHttpClient.cpp
+++ b/ixwebsocket/IXHttpClient.cpp
@@ -176,7 +176,7 @@ namespace ix
         ss << "Host: " << host << "\r\n";
 
 #ifdef IXWEBSOCKET_USE_ZLIB
-        if (args->compress)
+        if (args->compress && !args->onChunkCallback)
         {
             ss << "Accept-Encoding: gzip"
                << "\r\n";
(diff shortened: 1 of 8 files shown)


ixwebsocket/IXHttpClient.cpp
```diff
@@ -406,10 +406,10 @@ namespace ix
             ss << headers["Content-Length"];
             ss >> contentLength;
```
bsergean, line 407: Can you keep the payload.reserve call ? It will be better in the general case.

bsergean:
Thanks for the contribution. Can you mention this in the http client doc (in the docs subfolder) ?

ixwebsocket/IXHttpClient.cpp
natano: Hey! I moved the `reserve` call a couple of lines down (line 429, after the `readBytes` call), because it shouldn't be called for the case where a chunk callback is used. Is that good enough, or do you want it to happen before `readBytes`?

natano:
> Thanks for the contribution. Can you mention this in the http client doc (in the docs subfolder) ?

I added some documentation to `docs/usage.md`. Is that the right file?

ixwebsocket/IXHttpClient.cpp
bsergean: You are right that it does not matter, it just has to be done right before we write into the buffer, so the place where you moved it is fine.

bsergean:
Yes usage.md is the right place.

If you are gonna add doc for sure I can merge this, or maybe we just wait for the doc to be ready and merge everything at once.

natano:
I already added docs to [usage.md](https://github.com/machinezone/IXWebSocket/pull/353/files#diff-72376d0b487d7231cf31959bf266f6aea098e899743bae02e36d176cef4db476) :) Is there something else I should describe?

bsergean:
Ah sorry I missed that ... things look good enough, thanks for tweaking the doc and the unittest.

Issue #354, opened 2022-01-31
no way for websocket server to include Sec-WebSocket-Protocol header


sethalves:
If a client uses a sub-protocol while connecting to a websocket server, there's doesn't seem to be any way for the server to answer the client with a matching Sec-WebSocket-Protocol.

bsergean:
This header should be computed by the server, and be transparent, you don't need to specify it.

bsergean:
Ah sorry I got confused with the Sec-WebSocket-Accept header.

bsergean:
iirc, there is sub-protocol support for the client, but not for the server.

bsergean:
### Subprotocols

You can specify subprotocols to be set during the WebSocket handshake. For more info you can refer to [this doc](https://hpbn.co/websocket/#subprotocol-negotiation).

```cpp
webSocket.addSubprotocol("appProtocol-v1");
webSocket.addSubprotocol("appProtocol-v2");
```

The protocol that the server did accept is available in the open info `protocol` field.

```cpp
std::cout << "protocol: " << msg->openInfo.protocol << std::endl;
```

bsergean:
If you are trying to add this yourself, maybe it should go in

```
    WebSocketInitResult WebSocket::connectToSocket(std::unique_ptr<Socket> socket,
                                                   int timeoutSecs,
                                                   bool enablePerMessageDeflate)
```

for symmetry with 

```
    WebSocketInitResult WebSocket::connect(int timeoutSecs)
```

which create some headers, and pass them down with

```
        WebSocketInitResult status = _ws.connectToUrl(_url, headers, timeoutSecs);
```

sethalves:
I started down this path, but the problem is that the subprotocol needs to be set during the on-connection callback but before the on-message callback.  The program using the IX interface doesn't have access to the request headers during this time.  So, if I assume I know what the clients subprotocol will be, I *can* set it there, but it didn't seem like the right solution.

sethalves:
this is what I tried: https://github.com/machinezone/IXWebSocket/pull/355

bsergean:
Yes I had forgotten about how things work, what would be good would be to have a hook that happen before the websocket connection is handled, during the handshake. With an onHandshake callback maybe.

Other websockets libraries do that, and for example you can handle some basic authentication there, or reject some clients based on the sub-protocol etc...

dylan-hart:
[...]

PR #355, opened 2022-02-01
[do not merge] add a way for websocket server to include a sub-protocol in its response to the client

sethalves:


sethalves:
This isn't a good solution -- I've opened this PR to explain a failed attempt.

https://github.com/machinezone/IXWebSocket/issues/354

PR #356, opened 2022-02-08
Update IXSocketConnect.cpp

flagarde:
Part of the code is never executed
warning raised by clang

### DIFF
--- a/ixwebsocket/IXSocketConnect.cpp
+++ b/ixwebsocket/IXSocketConnect.cpp
@@ -91,10 +91,6 @@ namespace ix
                 return -1;
             }
         }
-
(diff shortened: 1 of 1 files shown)


ahausladen:
That code was never reached since revision 97255fbd62bd6929adeb93ccebe88e1ee6f8395d (2019-01-03, "cancellation refactoring") when the "maxRetries" for-loop condition was removed. The timeout should be handled by isCancellationRequested(). At least it has the capability (makeCancellationRequestWithTimeout).

bsergean:
Thanks @ahausladen / so we can merge this then, and maybe update the comments.

Issue #352, opened 2022-01-29
Do not redefine 


Tanishq-Banyal:

bsergean:
Probably we need to start using std::size_t, the one from C++.

https://en.cppreference.com/w/cpp/types/size_t

Do you want to make a PR for this ? I don't have access to a windows box.

bsergean:
You will need to include this first before usage -> <cstddef>

bsergean:
@Tanishq-Banyal can you try the branch from @flagarde in his pull request ?

bsergean:
That code was merged, so you can try with the master branch.

Tanishq-Banyal:
Yes it seems to work fine now

PR #357, opened 2022-02-09
Fix for MINGW32 and clang on windows (#352)

flagarde:
Hi,
Thanks for this nice library.

Using `msys2 github workflow` msys2/setup-msys2@v2 Clang64 and MINGW32(#352) builds are failling.

These PR fix compilation for this workflow. (tests have not been run)

### DIFF
--- a/ixwebsocket/IXNetSystem.cpp
+++ b/ixwebsocket/IXNetSystem.cpp
@@ -8,6 +8,12 @@
 #include <cstdint>
 #include <cstdio>
 #ifdef _WIN32
+#ifndef EAFNOSUPPORT
+  #define EAFNOSUPPORT 102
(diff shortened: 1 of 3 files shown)


bsergean:
The test were run eventually, not sure why.

bsergean:
Thanks for the good words, I hope this library help you do useful things.

Do you want to add one github action / workflow for those other compilers ? If they aren't used as much not sure it's worth it but we could. I don't have a strong opinion.

flagarde:
I could do a PR if you want. I have some workflows for this `https://github.com/yaodaq/YAODAQ/blob/main/.github/workflows/Windows-MSYS2.yml`. I could change it to follow your style.

PR #369, opened 2022-03-04
Install shared library to bin directory on windows

ardanew:
put ixwebsocket.dll near ws.exe on windows to be able to run it


bsergean:
Can we do it the other way around, and make the ws install program install itself in the ixwebsocket library ?

I'm afraid this will break something (vcpkg or conan packaging).

ardanew:
> Can we do it the other way around, and make the ws install program install itself in the ixwebsocket library ?
install ws.exe to install_prefix/lib? it will require ixwebsocket.dll to run. at the moment when building shared (dll) library it isn't installed anywhere.

> I'm afraid this will break something (vcpkg or conan packaging)
vcpkg downloads static version of the library, conan places ixwebsocket.dll in /bin. don't know if the PR will break it.

bsergean:
Sorry for the slow responses. Currently the project works ok with conan and vcpkg, so I'm not too willing to changes things, just for the ws tool which isn't a library.

Maybe it's fine because building ws is optional, and enabled through cmake.

PR #370, opened 2022-03-04
Export static symbols when building ws with shared library

ardanew:
CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS from f53b2f8 doesn't export static class members, so when building ws.exe linked against shared ixwebsocket library VS displays errors 'LNK2019 symbol not found'

the solution:
define IXWEBSOCKET_EXPORT to __declspec(dllexport) when building windows dynamic library, to __declspec(dllimport) when using it, to nothing in other cases (static or not windows). then apply this define to all the static members in all .h files
like `IXWEBSOCKET_EXPORT const static bool kDefaultEnablePong;`


bsergean:
Thanks @ardanew !

bsergean:
It looks like this code change broke the mingw build -> https://github.com/machinezone/IXWebSocket/runs/5731696045?check_suite_focus=true

PR #384, opened 2022-04-12
Feature/vcpkg zlib dep fix

bsergean:


flagarde:
Hi,

I think this link would help https://cmake.org/cmake/help/latest/guide/importing-exporting/index.html#id8

Most project Put some MyWonerfulProjectConfig.cmake in the cmake folder and add the generated targets to it. Maybe a example like this could be usefull : https://github.com/facebook/folly/blob/main/CMake/folly-config.cmake.in . This part of CMake is very hard to understand. I'm always very confused

PR #389, opened 2022-04-26
Fix `Sec-WebSocket-Key` to contain valid Base64.

rsmmr:
The generated header only "looked like" Base64, but if the other side actually tried to decode it as such, it could fail. This change fixes that to always generate a valid Base64 value.

This adds external code to do the base64 encoding. Not sure where you'd like to locate that inside the source tree, happy to change that to a different location, just let me know.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,7 @@ set( IXWEBSOCKET_SOURCES
 )
 
 set( IXWEBSOCKET_HEADERS
+    ixwebsocket/IXBase64.h
     ixwebsocket/IXBench.h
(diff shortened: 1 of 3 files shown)


bsergean:
Thanks !

Can you call this new file IXBase64.h, and put it in the same folder as IXWebsocket.h ?

bsergean:
Also the header file needs to be referenced here -> https://github.com/machinezone/IXWebSocket/blob/master/CMakeLists.txt#L117

rsmmr:
I updated the PR with those changes.

bsergean:
Thanks @rsmmr ! I just merged this.

PR #391, opened 2022-05-02
Fix host string length when passed to OpenSSL.

maxweisel:
The current implementation sets an empty host name with OpenSSL and is always rejected by BoringSSL: https://boringssl.googlesource.com/boringssl/+/master/crypto/x509/x509_vpm.c#93


maxweisel:
@bsergean not sure why the windows test suite is still running 4+ hours later, but it looks like this change passed on all of the other platforms.

bsergean:
Good catch, thanks. So this should only affect pre openssl-1.1 version, that shouldn't be the majority.

Not sure why the windows builder was hosed. TLS isn't activated in that builder so your change can't possibly affect it.

bsergean:
Ah I got this wrong it's for SNI, and openssl >= 1.1 (boringssl must set its version to act like that).

PR #399, opened 2022-06-30
Add option to disable hostname check

rsmmr:
This adds a new option that disables the hostname check for received certificates. This is useful in setups where one knows that a host's name won't align with the certificate.

The option is only implemented for OpenSSL right now, as that's where I need it. I can't easily test the other SSL implementations, so I figured I submit this as is for now to see what you think about it.


rsmmr:
Thanks for the feedback. I'll take a stab at it, but it might take me a bit.

rsmmr:
Updated this as requested, take a look. I added support to Apple SSL as well, but can't test that one because my tests spawns up an SSL server, which seems unsupported with Apple SSL. (Couldn't get the `badssl.com` endpoint to work and this might be the better test anyways.)

I added a couple of maintainance commits at the beginning, let me know if you'd rather not have them included.

bsergean:
Sorry I missed your update, will take a look soon.

bsergean (approved):

bsergean:
Thanks !

PR #401, opened 2022-07-15
Fix spelling mistake in docs

SeizureSaladd:
I am really new to CMake and C++ and I am no expert but I think it should say NOTFOUND instead of NOTFOUNS.

(Bit of a side note which has nothing to do with this pull request but I thought I might as well mention, I also installed IXWebSocket using vcpkg and want to use it in a CMake project but IXWEBSOCKET_INCLUDE_DIR and IXWEBSOCKET_LIBRARY are both NOTFOUND)


bsergean:
Thanks for the PR.

PR #405, opened 2022-07-25
add Content-Type support

mochazi:
image

image

image

The old version of `IXWebSocket` did not correctly handle the front-end static file type, causing the front-end page rendering to fail.

The browser needs to recognize the `Content-Type` parameter to determine the file type for rendering, so it is necessary to manually declare the `Content-Type` parameter.


bsergean:
Thanks for the PR. 

Can you update the formatting to match the project style ?

PR #423, opened 2022-12-21
Fix macos link error

crjc:
Linking to this lib isn't necessary on macOS and was preventing me from building.


CMakeLists.txt
```diff
@@ -251,10 +251,12 @@ if (WIN32)
   endif()
 endif()
```
bsergean, line 253: Can you make it

if(APPLE)
... your new line / not even sure it's needed.
elsif()
... the old stuff

crjc: Sorry, you're right - it wasn't needed. I've updated it now.

bsergean:
lgtm, thanks !

PR #426, opened 2023-01-03
Prevent deadlock when the server is stopping

itytophile:
During my tests, I like to start and stop a WebSockets server inside an infinite loop. However, sometimes the loop stops.

I think the problem lies in `_conditionVariableGC`. We use this condition variable to notify the `runGC` thread that something has to be "garbage collected". I think the deadlock happens when we notify the condition variable when the `runGC` is not yet waiting for the condition variable. To prevent this, I use the boolean member `_canContinueGC` to "buffer" the notification. If we want to notify `runGC` while it's not waiting, we can just set `_canContinueGC` to true and `runGC` will know that we called `notify_one`. After `runGC` has checked `_canContinueGC`, it resets it to `false`.

You can easily reproduce the deadlock with this code:
```cpp
int port = getFreePort();
int count = 0;
while(true) {
    std::cout << count++ << std::endl;
    ix::WebSocketServer server(port);
    REQUIRE(server.listen().first);
    server.start();
}
```

### DIFF
--- a/ixwebsocket/IXSocketServer.cpp
+++ b/ixwebsocket/IXSocketServer.cpp
@@ -219,6 +219,10 @@ namespace ix
         if (_gcThread.joinable())
         {
             _stopGc = true;
+            {
+                std::lock_guard<std::mutex> lock{ _conditionVariableMutexGC };
+                _canContinueGC = true;
+            }
             _conditionVariableGC.notify_one();
             _gcThread.join();
             _stopGc = false;
@@ -451,7 +455,10 @@ namespace ix
             if (!_stopGc)
             {
                 std::unique_lock<std::mutex> lock(_conditionVariableMutexGC);
-                _conditionVariableGC.wait(lock);
+                if(!_canContinueGC) {
+                    _conditionVariableGC.wait(lock, [this]{ return _canContinueGC; });
+                }
+                _canContinueGC = false;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


bsergean:
Can you make your new boolean an std::atomic, so that we don't have to use a mutex ?

itytophile:
From https://en.cppreference.com/w/cpp/thread/condition_variable
> Even if the shared variable is atomic, it must be modified while owning the mutex to correctly publish the modification to the waiting thread.

You can check cppreference's example too. If you wish, I can add a comment to tell the user that he must lock the mutex before accessibg the boolean.

bsergean:
Thanks for the pointers.

I don’t have much time to seriously look into this. Did you check that we follow recommended guidelines for the condition variable usage (by using mutex correctly)

itytophile:
I've just checked. In my understanding, you don't use condition variables to wait for a variable to be equal to a specific value. You use condition variables to tell to another thread "there is something new". So it isn't really the same problem that the mutexes solve (in my opinion).
This "thread notifying" pattern is used in:
- `IXHttpClient`, to ask a thread to perform a task or stop the thread. There are a lot of `if (_stop) return` in the same function, that's really strange.
- `IXWebsocket`, to check new connections or stop the thread.
- `IXSocketServer`, to ask the GC to do something (fixed by this PR I hope) or to wait for the server to stop

Channels would be a great fit for this use case imo (like https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html in Rust) but I don't know if there is a good C++ lib in the wild. In any case, I think that the main problem is the same one that is solved by this PR: it is possible to notify the condition variable before the other thread is waiting, thus an event is skipped.

It is possible that I'm missing something so be careful!

PR #427, opened 2023-01-03
HTTP should contain port in 'Host' header

Vol-Alex:
For connection like "http://localhost:8180/" we need send the host name with the port. For more info see https://www.rfc-editor.org/rfc/rfc9110.html#name-host-and-authority


bsergean:
Could we only do that if the port is not 80 or 443 ?
I wonder what curl does.

bsergean:
Thanks for the PR btw.

bsergean:
Thanks @Vol-Alex !

PR #443, opened 2023-02-19
Fix compilation on GCC-13

winterheart:
GCC-13 changes internal cstdint includes, and now files that uses standart integer types should directly include cstdint header.

See: https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes
Bug: https://bugs.gentoo.org/865117
Bug: https://bugs.gentoo.org/895440


bsergean:
Thanks for the pr !

PR #445, opened 2023-02-28
Check ip adress

peterphonic:
IXWebSocket can now validate certificate with not only hostname, but with IP addresses as well. @bsergean


ixwebsocket/IXSocketOpenSSL.cpp
```diff
@@ -16,17 +16,23 @@
...
 #ifdef _WIN32
 #include <shlwapi.h>
+//To avoid the problem with X509_NAME collision between macro in wincrypt.h and x509v3.h
+//It has to be undef before x509v3.h and after wincrypt.h
+#undef X509_NAME
 #else
 #include <fnmatch.h>
 #endif
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
+
```
bsergean, line 25: I'm pretty positive that old openssl installation will not have this x509v3.h header, so we should keep this OPENSSL_VERSION_NUMBER check or we will lose compatibility with old openssl.

ixwebsocket/IXSocketOpenSSL.cpp
```diff
@@ -77,6 +83,43 @@ namespace
 
         return true;
     }
+
```
bsergean, line 86: Can you add unittest for this new function ?

bsergean (commented):
Thanks for contributing. See inline comments.

ixwebsocket/IXSocketOpenSSL.cpp
peterphonic: I am just a little bit confused here.  They way I read this is "`if version < 0X10100000L`" we **do** include `x509v3.h`.  They way you explaining sound like the opposite, with the line I removed,  `x509v3.h` is not included.  But then, I would expect this instead "`if version > 0X10100000L`"

I am using OpenSSL 3.0.7, and if I keep the line above, I don't have access to define from  `x509v3.h`.  So, I would suggest this : 
```
#if OPENSSL_VERSION_NUMBER > 0x10100000L
#include "openssl/x509v3.h"
#endif
```
So, the include file will be there only on more recent version of OpenSSL.

ixwebsocket/IXSocketOpenSSL.cpp
peterphonic: I guess so, where should I add this?  Is is the reason I got message that test failed under linux, linux_asan, and mac_tsan... ?

bsergean:
There is a tests subfolder at the top, you can create a new file in there or add to an existing file.

test/IXSocketTest.cpp
```diff
@@ -95,4 +95,19 @@ TEST_CASE("socket", "[socket]")
         testSocket(host, port, request, socket, expectedStatus, timeoutSecs);
     }
 #endif
+
+TEST_CASE("isValidIpAddress") {
```
peterphonic, line 99: I don't know if this works, I am not able to compile the test folder under windows.  I also tried under Centos7 with build_linux.sh, but some files are missing.  So, I don't know if I am supposed to build....

bsergean:
All the PR checks have failed.

bsergean:
Hey there, sorry this got forgotten about. A fix for that just landed in #588.

If we need more support we can reopen.

PR #450, opened 2023-03-15
fix incorrect closures with code 1011 Internal error

### LINKED ISSUE #417: close code 1011 (internal error) behavior
im not entirely sure whats going on here, but it seems incorrect. having the server close the connection (the message sent in the callback will make the server close) results in a `1011 Internal error` rather frequently. the subsequent closeInfo then appears to have the corret code but an empty reason. i made an equivalent program in python and it consistently returns `4003 Not authenticated.` also not sure if its related, but i have to .stop() even after the websocket is `CLOSED` or else .start() seems to do nothing (i never get an open frame). reconstructing the websocket within the loop also seems to have the same behavior except the reason field doesnt go empty.

```cpp
#include <ixwebsocket/IXNetSystem.h>
#include <ixwebsocket/IXWebSocket.h>
#include <iostream>

#pragma comment(lib, "crypt32.lib")

int main() {
    ix::initNetSystem();

    ix::WebSocket webSocket;
    webSocket.disableAutomaticReconnection();
    webSocket.setUrl("wss://newark3688.discord.media:443/?v=7");
    webSocket.setOnMessageCallback([&](const ix::WebSocketMessagePtr &msg) {
        if (msg->type == ix::WebSocketMessageType::Message) {
            std::cout << "received message: " << msg->str << std::endl;
            webSocket.send(R"({"d":1668653651493,"op":3})");
        } else if (msg->type == ix::WebSocketMessageType::Open) {
            std::cout << "Connection established" << std::endl;
        } else if (msg->type == ix::WebSocketMessageType::Error) {
            std::cout << "Connection error: " << msg->errorInfo.reason << std::endl;
        } else if (msg->type == ix::WebSocketMessageType::Close) {
            std::cout << "Connection closed: " << msg->closeInfo.code << " (" << msg->closeInfo.reason << ")";
            if (msg->closeInfo.remote) {
                std::cout << " Remote";
            }
            std::cout << std::endl;
        }
    });

    while (true) {
        std::cout << webSocket.readyStateToString(webSocket.getReadyState()) << std::endl;
        webSocket.stop();
        webSocket.start();
        Sleep(1000);
    }

    return 0;
}
```

```
CLOSED
Connection established
received message: {"op":8,"d":{"v":7,"heartbeat_interval":13750}}
Connection closed: 4003 (Not authenticated.) Remote
CLOSED
Connection established
received message: {"op":8,"d":{"v":7,"heartbeat_interval":13750}}
Connection closed: 4003 (Not authenticated.) Remote
CLOSED
Connection established
received message: {"op":8,"d":{"v":7,"heartbeat_interval":13750}}
Connection closed: 4003 (Not authenticated.) Remote
CLOSED
Connection established
received message: {"op":8,"d":{"v":7,"heartbeat_interval":13750}}
Connection closed: 1011 (Internal error)
CLOSED
Connection established
received message: {"op":8,"d":{"v":7,"heartbeat_interval":13750}}
Connection closed: 4003 () Remote
CLOSED
Connection established
received message: {"op":8,"d":{"v":7,"heartbeat_interval":13750}}
Connection closed: 4003 (Not authenticated.) Remote
```

also, the loop is just for testing. you could just .start() and Sleep once and rerun the program and the internal error still shows up

ouwou:
closes #417

to re-summarize:

if a websocket server sends a close and immediately drops the connection, then ixwebsocket will close with "1011 Internal error" since it fails to reply with a close frame of its own. im not super sure how the connection is closing that causes it to error and im struggling to replicate it myself, but the code in the issue still causes it. [this](https://github.com/openssl/openssl/blob/19cc035b6c6f2283573d29c7ea7f7d675cf750ce/ssl/record/rec_layer_s3.c#L320-L321) is where openssl errors, but it happens with mbedtls as well

https://github.com/machinezone/IXWebSocket/blob/dc7b986e1092e59195085b708033723f4912b98f/ixwebsocket/IXWebSocketTransport.cpp#L699-L709
https://github.com/machinezone/IXWebSocket/blob/dc7b986e1092e59195085b708033723f4912b98f/ixwebsocket/IXWebSocketTransport.cpp#L1072-L1077

this pr makes it so the websocket goes into CLOSING just before replying, and checking if the state is CLOSING before trying to close again

some other things worth considering:
* would `IXWebSocket::close` still work if the send failed?
* the close reason seems to sometimes be empty for the close *after* the one that would cause ixwebsocket to give a 1011
* the ReadyState of the socket from within the user-defined callback will be CLOSING now instead of OPEN

you also mentioned the closing test (IXWebSocketCloseTest) which i was able to run on both mingw64 and msvc and it seems to have run and passed just fine, so maybe i can enable that as part of this pr?

### DIFF
--- a/ixwebsocket/IXWebSocketTransport.cpp
+++ b/ixwebsocket/IXWebSocketTransport.cpp
@@ -700,6 +700,7 @@ namespace ix
                 if (_readyState != ReadyState::CLOSING)
                 {
                     // send back the CLOSE frame
+                    setReadyState(ReadyState::CLOSING);
                     sendCloseFrame(code, reason);
 
                     wakeUpFromPoll(SelectInterrupt::kCloseRequest);
@@ -1072,7 +1073,10 @@ namespace ix
             else if (ret <= 0)
             {
                 closeSocket();
-                setReadyState(ReadyState::CLOSED);
+                if (_readyState != ReadyState::CLOSING)
+                {
+                    setReadyState(ReadyState::CLOSED);
+                }
                 return false;
             }
             else
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -24,14 +24,13 @@ set (TEST_TARGET_NAMES
   # IXWebSocketBroadcastTest ## FIXME was depending on cobra / take a broadcast server from ws
   IXStrCaseCompareTest
   IXExponentialBackoffTest
+  IXWebSocketCloseTest
 )
 
 # Some unittest don't work on windows yet
 # Windows without TLS does not have hmac yet
 if (UNIX)
   list(APPEND TEST_TARGET_NAMES
-    IXWebSocketCloseTest
-
     # Fail on Windows in CI probably because the pathing is wrong and
     # some resource files cannot be found
     IXHttpServerTest


bsergean:
Sorry I missed this. Yes could you try to enable that other test ?

We will see if the PR are happy with it, they run on all platform.

What's a bit concerning is the tsan failures, but those can fail randomly so maybe it's a red herring.

ouwou:
yeah, i understand your caution which is also why i didnt open a PR right away. is there something i should do to help with autobahn? im not really familiar with autobahn so i dont know how i could go about handling that.

bsergean:
I haven’t done that in a while, but here are some notes I took back then.

    //
    // Autobahn test suite
    //
    // 1. First you need to generate a config file in a config folder,
    //    which can use a white list of test to execute (with globbing),
    //    or a black list of tests to ignore
    //
    // config/fuzzingserver.json
    // {
    //     "url": "ws://127.0.0.1:9001",
    //     "outdir": "./reports/clients",
    //     "cases": ["2.*"],
    //     "exclude-cases": [
    //     ],
    //     "exclude-agent-cases": {}
    // }
    //
    //
    // 2 Run the test server (using docker)
    // docker run -it --rm -v "${PWD}/config:/config" -v "${PWD}/reports:/reports" -p 9001:9001
    // --name fuzzingserver crossbario/autobahn-testsuite
    //
    // 3. Run this command
    //    ws autobahn -q --url ws://localhost:9001
    //
    // 4. A HTML report will be generated, you can inspect it to see if you are compliant or not
    //

ouwou:
got it to run and every case has the same result as [here](https://bsergean.github.io/autobahn/reports/clients/index.html). i think ill try to do some testing of my own to check what happens i.e. when calling `close()` and the connection drops out since thats where i expect side effects to show up if there are any

bsergean:
Thanks for all the work @ouwou, I just merged your change, we see if people complains.

The only problem would be if autobahn output can differ depending on the platform.

PR #460, opened 2023-04-26
use deque for _rxbuf to avoid the vector erase() moving too much memo…

cvoica:
testing to see if this solution will work

### DIFF
--- a/ixwebsocket/IXWebSocketTransport.cpp
+++ b/ixwebsocket/IXWebSocketTransport.cpp
@@ -458,8 +458,9 @@ namespace ix
         while (true)
         {
             wsheader_type ws;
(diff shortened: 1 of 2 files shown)


bsergean:
There were a lot of failures on the PR checks.

cvoica:
I work on Linux and I'm happy yo see that at least those tests were fine as I currently use it without any issues on alinux.

After a brief look at Windows logs errors, like TLS support, I remeber this was a compilation flag missing in Linux too.

I will have an in-depth look tomorrow.

bsergean:
Sounds good no rush.

bsergean:
hey let's close this, it is pretty inactive.

PR #465, opened 2023-05-07
Use cmake cache variables to customize

hotwatermorning:
Hi.

I want to use IXWebSocket library with C++17 project but it seems that IXWebSocket can be built only for C++11.

Thus I modified CMakeLists.txt to use cache variables to enable customizing build settings outside of the file.
(And fixed CXX_STANDARD_REQUIRED usage. It should be CMAKE_CXX_STANDARD_REQUIRED because CXX_STANDARD_REQUIRED is a property.
https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html
)

Thanks.


rh101:
How about this instead?
```
set(_IXWEBSOCKET_MIN_CXX_STD 11)
if (NOT DEFINED CMAKE_CXX_STANDARD)
    set(CMAKE_CXX_STANDARD ${_IXWEBSOCKET_MIN_CXX_STD})
endif()
if (CMAKE_CXX_STANDARD LESS ${_IXWEBSOCKET_MIN_CXX_STD})
    message(FATAL_ERROR "IXWebsocket requires c++ std >= ${_IXWEBSOCKET_MIN_CXX_STD}")
endif()

if(NOT DEFINED CMAKE_CXX_STANDARD_REQUIRED)
    set(CMAKE_CXX_STANDARD_REQUIRED ON)
endif()

if(NOT DEFINED CMAKE_CXX_EXTENSIONS)
    set(CMAKE_CXX_EXTENSIONS OFF)
endif()
```

PR #467, opened 2023-05-11
Fix version in CMakeLists.txt

lanthora:


flagarde:
Maybe this version infos could be added inside the `project` command for simplicity ?

lanthora:
> Maybe this version infos could be added inside the `project` command for simplicity ?

I don't know what the "`project` command" means.

Usually Linux users use the following command to compile and install:

```bash
cmake ../
make
make install
```

Do you mean to specify the version number through the parameter when executing the cmake command?

```bash
cmake -DVERSION=x.x.x ../
```

I don't think it's a good idea to let the user decide the version number.

flagarde:
I mean the project command inside CMakeLists can have a version https://cmake.org/cmake/help/latest/command/project.html

Which can be iseful as it is easily to see and have parameter dedicated by CMAke for this usage

lanthora:
It looks ok now

bsergean (approved):

lanthora:
@bsergean IXHttpClientTest always fails recently, the master branch only modified the README file when all CIs failed last time. It may be a problem with the test server

PR #468, opened 2023-05-20
Fix compilation on Windows

N00byKing:
Was commented out recently, but identifier is still used a few lines below.


N00byKing:
Superseded by #471 for some reason.

PR #478, opened 2023-07-19
Fix server empty thread name

lanthora:
setThreadName in WebSocket::run will overwrite the previously set Srv::ws:, and getUrl is empty at this time, resulting in an empty thread name. For the client, 16 characters are not enough to represent very valid information in Linux. So delete setThreadName in WebSocket::run


bsergean:
16 chars is not a ton but is useful still, maybe there could be an option to disable this, and we would default it to false to keep the old behavior.

lanthora:
@bsergean Fix the problem that the server thread name is empty, and keep the old behavior. Add option to disable automatic update of thread name.

bsergean:
thanks !

Issue #474, opened 2023-06-26
the client did not receive the close event


engun:
System: Windows 11 64-bit.

This lib is great. Although there were some issues encountered during use

When the server is abnormally shut down, the client did not receive the close event. It seems that the FD_CLOSE event should be taken into account in `WSAWaitForMultipleEvents` in `IXNetSystem::poll`.

bsergean:
Thanks for the good words, it's possible there are some holes in the IXNetSystem::poll.

If you want to submit a PR we can try to evaluate it and merge it.

I code on macOS and Linux, so I'm not the best person to fix advanced windows problem sadly.

dacap:
It's probable that there is a missing `wakeUpFromPoll()` call from `WebSocketTransport::close()` to unlock the `WebSocket::stop()` function which is waiting for a `_thread.join()` (this thread is running `WebSocket::run()` which calls `WebSocketTransport::poll()` that can be waken up only if `wakeUpFromPoll()` is called again):

```diff
diff --git a/ixwebsocket/IXWebSocketTransport.cpp b/ixwebsocket/IXWebSocketTransport.cpp
index ec25465..8049fb6 100644
--- a/ixwebsocket/IXWebSocketTransport.cpp
+++ b/ixwebsocket/IXWebSocketTransport.cpp
@@ -1174,7 +1174,10 @@ namespace ix
     {
         _requestInitCancellation = true;
 
-        if (_readyState == ReadyState::CLOSING || _readyState == ReadyState::CLOSED) return;
+        if (_readyState == ReadyState::CLOSING || _readyState == ReadyState::CLOSED) {
+            wakeUpFromPoll(SelectInterrupt::kSendRequest);
+            return;
+        }
 
         if (closeWireSize == 0)
         {
```

bsergean:
@dacap If you make a PR for this change we can give it a try on CI.

Longwater1234:
This problem still persists even with latest version. Only on Windows. On MacOS it alright.

bsergean:
Might be fixed with https://github.com/machinezone/IXWebSocket/pull/548

PR #481, opened 2023-08-23
Fix hanging of WebSocket::stop() waiting for its thread to join

dacap:
This might happen in some special cases where `WebSocketTransport::sendOnSocket()` fails, closes the socket, and set the ready state to CLOSED, but `WebSocket::run()` stills wait the interrupt event to happen.

Possibly related to #474.


dacap:
I've just updated the PR to match the `if (...) { ... }` style.

bsergean:
looking good ci wise, re-ran them.

bsergean:
thanks @dacap

PR #482, opened 2023-08-25
Fix MinGW build warning

lanthora:
When using the compiler options -Wall and -Werror, warnings caused by type mismatches are treated as errors, leading to compilation failure.


bsergean:
thanks for the fix @lanthora

PR #483, opened 2023-08-25
Fix #286 - http response headers overwritten with request headers

glenne:
Not sure why it was copying request headers to the response so I took that out altogether.   Response headers should be determined by the server and not parroted from the client.   For example, Accept-Encoding and Content-Encoding should be set by the server and not the client as it was before.  With removal of the blind copy of request headers, I added explicit setting of Content-Encoding for the gzip ifdef case and added a test to verify it was set.


bsergean:
looking good, thanks @glenne

PR #487, opened 2023-10-12
Allow building without rtti

arenevier:
Since factory returns a ProxyConnectionState, setOnConnectionCallback will be a ProxyConnectionState. The code already makes that assumption, since it does not check of state return value. Using a static_pointer_cast will allow the library to be build with rtti.


bsergean:
Thanks for contributing !

PR #488, opened 2023-10-12
allow building when cpp exceptions are disabled

arenevier:
IXWebSocket needs exceptions support because of the use of stoi. In order to build when cpp exceptions are disabled, we can use strtol when __cpp_exceptions is not set

### DIFF
--- a/ixwebsocket/IXHttp.cpp
+++ b/ixwebsocket/IXHttp.cpp
@@ -133,6 +133,7 @@ namespace ix
         if (headers.find("Content-Length") != headers.end())
         {
             int contentLength = 0;
+#ifdef __cpp_exceptions
             try
             {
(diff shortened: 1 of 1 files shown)


bsergean:
Windows / gcc is not happy. Can we switch to your new code all the time ?

arenevier:
Are you talking about the build error? 

     | Cannot find path
     | 'C:\ProgramData\chocolatey\lib\mingw\tools\install\mingw64\x86_64-w64-mingw32\lib\libpthread.dll.a' because it
     | does not exist.
     
I'm not sure how it can be related to this change.

arenevier:
Anyway, I have created an  alternative pull request with only the exception safe codepath: https://github.com/machinezone/IXWebSocket/pull/489

PR #489, opened 2023-10-13
allow building when cpp exceptions are disabled

arenevier:
IXWebSocket needs exceptions support because of the use of stoi. In order to build when cpp exceptions are disabled, we can use strtol.

### DIFF
--- a/ixwebsocket/IXHttp.cpp
+++ b/ixwebsocket/IXHttp.cpp
@@ -133,16 +133,20 @@ namespace ix
         if (headers.find("Content-Length") != headers.end())
         {
             int contentLength = 0;
-            try
             {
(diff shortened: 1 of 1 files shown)


ixwebsocket/IXHttp.cpp
```diff
@@ -133,16 +133,20 @@ namespace ix
         if (headers.find("Content-Length") != headers.end())
         {
             int contentLength = 0;
```
bsergean, line 135: do we need this nested scope (open/close curly braces ?)

bsergean:
Can you rebase master ? I just disabled the windows gcc thing.

bsergean:
I think it's fine as is, thanks for the contribution.

PR #493, opened 2023-11-15
Fix for C++/WinRT compile issue

### LINKED ISSUE #492: UWP App undefined method PathMatchSpecA
image

`PathMatchSpecA` is in `shlwapi.h`, but that entire code is disabled when compiling a Universal Windows application (WinRT/C++):

image

Due to this check:
image

I noticed that `SocketOpenSSL::checkHost()` is only used if `OPENSSL_VERSION_NUMBER < 0x10100000L`, and it's a private method, so not called from anywhere outside of `SocketOpenSSL`.  In my case it's also not being called at all given that the OpenSSL version number is greater than `0x10100000L`, so as a work-around, I've just changed it to this:
```
    /**
     * Check whether a hostname matches a pattern
     */
    bool SocketOpenSSL::checkHost(const std::string& host, const char* pattern)
    {
#if OPENSSL_VERSION_NUMBER >= 0x10100000L
        return true;
#else

#ifdef _WIN32
        return PathMatchSpecA(host.c_str(), pattern);
#else
        return fnmatch(pattern, host.c_str(), 0) != FNM_NOMATCH;
#endif

#endif
    }
```

Can you see any issue with doing this?

rh101:
Closes #492

### DIFF
--- a/ixwebsocket/IXSocketOpenSSL.cpp
+++ b/ixwebsocket/IXSocketOpenSSL.cpp
@@ -26,6 +26,7 @@
 
 #ifdef _WIN32
 // For manipulating the certificate store
+#include <windows.h>
 #include <wincrypt.h>
 #endif
 
@@ -293,10 +294,16 @@ namespace ix
      */
     bool SocketOpenSSL::checkHost(const std::string& host, const char* pattern)
     {
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+        return true;
+#else
+
 #ifdef _WIN32
         return PathMatchSpecA(host.c_str(), pattern);
 #else
         return fnmatch(pattern, host.c_str(), 0) != FNM_NOMATCH;
+#endif
+
 #endif
     }
 


bsergean:
If you have access to a mingw environment on Windows, it would be great if you could figure out that build issue.

rh101:
> If you have access to a mingw environment on Windows, it would be great if you could figure out that build issue.

I'll see what I can do.

rh101:
Are you certain that the error displayed in the [logs](https://github.com/machinezone/IXWebSocket/actions/runs/6873411557/job/18716750468?pr=493) has something to do with the changes made in this PR?

The specific error is this, which seems to come from the `https://github.com/egor-tensin/setup-mingw` script:
```
 Remove-Item: D:\a\_temp\14562f07-631d-4e82-afd4-223d04e7d198.ps1:140
Line |
 140 |          Remove-Item (Join-Path $mingw_lib 'libpthread.dll.a')
     |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot find path
     | 'C:\ProgramData\chocolatey\lib\mingw\tools\install\mingw64\x86_64-w64-mingw32\lib\libpthread.dll.a' because it
     | does not exist.
```

I installed mingw, and built ixwebsocket based on the workflow in this file: https://github.com/machinezone/IXWebSocket/actions/runs/6873411557/workflow?pr=493

No errors at all. Reverting the changes I made in this PR also resulted in the same output.  Also, the cmake command in that workflow doesn't enable SSL, so the changes I made are not part of the build.

bsergean:
Yes your PR did not break this, it was broken before.

I was asking if you'd be interested in helping fix the mingw build.

rh101:
> Yes your PR did not break this, it was broken before.
> 
> I was asking if you'd be interested in helping fix the mingw build.

My misunderstanding, but sure, I'll take a more detailed look into what may be causing that specific issue when I can.

rh101:
The issue is in `setup-mingw`, and it seems that this has already been reported, and a fix provided, but has not been merged in:

https://github.com/egor-tensin/setup-mingw/issues/15
https://github.com/egor-tensin/setup-mingw/pull/16

If you fork that repository, apply the fix, and map your workflow link to your own version, then it should work fine.

bsergean:
Thanks for researching this.

If you care and would like to make a PR for that it would be very much appreciated.

rh101:
> If you care and would like to make a PR for that it would be very much appreciated.

I can't create a PR unless you create a fork of this repository first: 
https://github.com/egor-tensin/setup-mingw

Once that repository is under the"machinezone" github account, then it's simply a matter of merging this commit with your version:
https://github.com/egor-tensin/setup-mingw/pull/16/files

Also, a PR would then be created against the IXWebSocket workflow file: 
https://github.com/machinezone/IXWebSocket/blob/master/.github/workflows/unittest_windows_gcc.yml

Changing this line:
`- uses: egor-tensin/setup-mingw@v2.2.0`

to something like this:
`- uses: machinezone/setup-mingw@your-patched-version`

bsergean:
Thanks for the help, I applied the fix in my own fork and now gcc build works on windows.

https://github.com/machinezone/IXWebSocket/pull/494

PR #497, opened 2024-01-18
Add ping interval to constructor params for WebSocketServer

dwymark-celestron:
I could not find any way to enable the ping heartbeat for all connections while configuring the WebSocketServer class. I added the parameter pingIntervalSeconds to the constructor to allow for this to be configured at construction time. Per my reading of the code and some local sanity checking, setting this parameter for all child connections enables the heartbeat feature.

If there was a better way for me to achieve this without the modification in this PR, please let me know and we can close. Otherwise, let me know if there are any further modifications you would like before we can merge this into master.


bsergean:
CR passed, do you mind checking if the documentation should be updated ? (under the docs folder) ?

dwymark-celestron:
I added a paragraph to the usage documentation. I couldn't find anywhere else relevant worth updating. Let me know if you have any other requests to move this PR forward.

bsergean (approved):

bsergean:
Thanks ! Just merged your change.

PR #500, opened 2024-02-16
Fix MbedTLS disconnect handling.

PaulWhitingS2:
Currently, the code spikes the CPU to 100% while waiting to receive a message on a broken connection when using MBedTLS.  It fails to detect the broken connection.  This pull request fixes that.

### DIFF
--- a/ixwebsocket/IXSocketMbedTLS.cpp
+++ b/ixwebsocket/IXSocketMbedTLS.cpp
@@ -352,6 +352,11 @@ namespace ix
                 return res;
             }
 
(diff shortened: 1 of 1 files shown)


CryptoManiac:
It's correct for the OpenSSL backend as well.

bsergean (approved):

bsergean:
Thanks @PaulWhitingS2 !

bsergean:
I should have run the PR check before I merged this, there was a PR check that started breaking after this it seems.

PR #502, opened 2024-02-23
Don't call close() on -1.

CryptoManiac:
Valgrind keeps complaining that close() on the invalid descriptor -1 is happening here. I suppose that close shouldn't be called when the descriptor value is known to be invalid. It's not a fatal error or something, but this practice is able to create a lot of flood in the logs.

### DIFF
--- a/ixwebsocket/IXSelectInterruptPipe.cpp
+++ b/ixwebsocket/IXSelectInterruptPipe.cpp
@@ -34,8 +34,12 @@ namespace ix
 
     SelectInterruptPipe::~SelectInterruptPipe()
     {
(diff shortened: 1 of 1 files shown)


bsergean:
Thanks !
I usually put the -1 on the right hand side but that's fine.

bsergean:
I had forgotten that Yoda denomination.

Could you flip it the normal way? I like the code to stay consistent with
the classical way.

CryptoManiac:
> I had forgotten that Yoda denomination. Could you flip it the normal way? I like the code to stay consistent with the classical way.

Sure, I'll revise this tomorrow I think.

PR #503, opened 2024-02-23
Don't read() from the invalid descriptor

CryptoManiac:
Having the invalid value of descriptor here might be considered weird, but it happens nevertheless, when you're DoSing the websocket endpoint using the 3G connection. It makes sense to return the generic error here instead of trying to read() from it.


bsergean:
Can you return -1 instead ?

cosminpolifronie:
@bsergean Can be closed as it is solved by #573

bsergean:
@cosminpolifronie you are right

PR #505, opened 2024-03-07
init simple debian build configs and bump version to 11.4.5

YangMame:
can you re-release the tag?


YangMame:
image
wrong version will case rpmlint check failure

PR #506, opened 2024-03-08
Release v11.4.5

lanthora:
Change soname to avoid conflicts with older versions on some systems.

https://github.com/machinezone/IXWebSocket/pull/505


lanthora:
It may be a better choice to modify the version number in cmakelists to 11.4.6 and add a new tag

PR #507, opened 2024-03-08
Support URLs with no slash before the question mark

giuspen:
A URL such wss://google.com?arg=value&arg2=value2 is also valid, this PR adds support for it


bsergean:
Thanks !

PR #512, opened 2024-04-19
IXSocket.h: add missing <sys/types.h> for macOS

barracuda156:
Fixes an error with undefined `ssize_t`


bsergean:
something is unhappy with one check

barracuda156:
Was it passing otherwise? It is strange if build is fine but tests fail because of adding a header. But I can try locally.

PR #514, opened 2024-04-21
Initialize the PSA Crypto API if requested

teejusb:
I have an emulator that makes use of IXWebSocket for networking under the hood [here](https://github.com/itgmania/itgmania).

When updating to the latest MbedTLS 3.6.0 LTS, I noticed that our SSL requests started failing, while non-SSL requests were still fine. After some digging, it seems like calling `psa_crypto_init()` somewhere in the code allowed it to start working again, which is what this PR suggests.

In the MbedTLS codebase ([here](https://github.com/Mbed-TLS/mbedtls/blob/development/docs/use-psa-crypto.md)), it says that by using `MBEDTLS_USE_PSA_CRYPTO` we promise to also call `psa_crypto_init()` before calling any PK, X.509 or TLS functions, so we control that using a compile time definition named `IXWEBSOCKET_MBEDTLS_USE_PSA_CRYPTO`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ If your company or project is using this library, feel free to open an issue or 
 - [Abaddon](https://github.com/uowuo/abaddon), An alternative Discord client made with C++/gtkmm 
 - [NovaCoin](https://github.com/novacoin-project/novacoin), a hybrid scrypt PoW + PoS based cryptocurrency.
 - [Candy](https://github.com/lanthora/candy), A WebSocket and TUN based VPN for Linux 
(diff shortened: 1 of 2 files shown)


bsergean:
Thanks !

bsergean:
Hey there's one regression do you know what could happen ?

One unittest now fail on CI for mbdedtls.

teejusb:
Strange. Let me take a look and see if I can figure it out. Since the code only looked for a not-previously-introduced DEFINE it should have been a no-op in general.

teejusb:
Created https://github.com/machinezone/IXWebSocket/pull/516!

PR #517, opened 2024-05-12
Fix cleanup and resource management in example

ghost:
Also add some more example code as well as refactoring of some of the existing codebase.


bsergean:
Can you restore the indentation to how it used to be ?
And just add the new CA block ?

PR #527, opened 2024-09-03
Fix memory leaks in SocketMbedTLS when closing.

plmzod:
Hello (and thanks for this library!), I've found two leaks when using sockets with SSL/TLS and MbedTLS.
One leak about the key, and one from the new PSA layer introduced with MbedTLS 3.6.0.
The leaks are detected on Windows with VLD and on Linux with ASAN.


bsergean:
Thanks for the fixes !

PR #530, opened 2024-10-30
Allow Host to be overridden in handshake headers

gordonread:
I had a situation with a thirdparty websocket server which was rejecting the upgrade to a websocket connection (in `IXWebSocketHandshake::clientHandshake(...)`) with a HTTP 400 (Bad Request) error because the thirdparty server didn't like the format of the "Host:" header being set to "stream.example.com:port", so, e.g.

```
GET /required/path HTTP:/1.1
Host: stream.example.com:1234
```

Failed with a HTTP 400 error code (when IXWebSocket expects a HTTP 101 Switching Protocols code), but the following without ":port":

```
GET /required/path HTTP:/1.1
Host: stream.example.com
```

Worked OK.

This small change allows the "Host" header to be specified in `ix::WebSocketHttpHeaders", e.g.

```
ix::WebSocket webSocket{};
ix::WebSocketHttpHeaders headers{};
headers["Host"] = "stream.example.com";
webSocket.SetExtraHeaders(headers);
```

Will give the desired result.

Added a few test cases.

### DIFF
--- a/ixwebsocket/IXWebSocketHandshake.cpp
+++ b/ixwebsocket/IXWebSocketHandshake.cpp
@@ -114,7 +114,10 @@ namespace ix
 
         std::stringstream ss;
         ss << "GET " << path << " HTTP/1.1\r\n";
-        ss << "Host: " << host << ":" << port << "\r\n";
+        if (extraHeaders.find("Host") == extraHeaders.end())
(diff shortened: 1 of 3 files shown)


bsergean:
Thanks !

PR #534, opened 2024-12-03
Fix: 'Content-Length' header check implicit cast

Rodrick136:
Heyo,
Thank you for the awesome library!

I fixed a mildly annoying compiler warning by making this a static cast instead of an implicit one.

I also rearranged some of the checks to add better error messages to the 'Content-Length' header check overall.

First ever contribution to a public repo, let me know if it checks out!


bsergean (approved):
Thanks man !

Issue #531, opened 2024-11-17
Add -Werror -Wshadow compiler flag


vaezim:
`-Werror -Wshadow` is a common flag in most projects. IXWebSocket  cannot be built with this flag.

bsergean:
Do you want to make a PR to fix this ?

vaezim:
Ok, I'll try.

vaezim:
Created the PR

PR #535, opened 2024-12-08
Add -Werror -Wshadow compiler flags to CMAKE_CXX_FLAGS (#531)

vaezim:


rh101:
@vaezim It looks like you accidentally change a code completely unrelated to the `-Werror -Wshadow` compiler flags.  There shouldn't be underscores at the start of all those variable names.

vaezim:
> @vaezim It looks like you accidentally changed code completely unrelated to the `-Werror -Wshadow` compiler flags. There shouldn't be underscores at the start of all those variable names.

Those variables were shadowed by another variable in the outer scopes. Either outer scope variable or local ones needed renaming. Do you have other solutions? we can clear the outer one and reuse it.

rh101:
> Those variables were shadowed by another variable in the outer scopes. Either outer scope variable or local ones needed renaming. Do you have other solutions? we can clear the outer one and reuse it.

If that's the case, then it would be best to rename them to something more appropriate, without the underscore prefix, since by convention that makes them appear as class member variables, which they are not.

vaezim:
Removed underscores and renamed according to scope context.

PR #542, opened 2025-02-12
removed default user-agent and origin headers

zxsleebu:
this is actually needed in some cases.
or you can make this behaviour optional, but i'm not really sure how it should look like


bsergean:
I think you can add an option for this.

PR #549, opened 2025-04-22
Improve documentation and add an ios build script

### LINKED ISSUE #546: Creating static libraries for a multi platform project
Hello,
we are developing a mobile game and would like to switch to IXWebSocket for handling socket connections. For this we would need static libraries for the android (armeabi-v7a, arm64-v8a & x86), iOS and mac. We are struggling a bit to compile all these from the project and wondered if we could get some assistance for this case. 
The project states it is compatible with android, iOS and mac so would it be possible to get some more detailed instructions for building the project for these platforms? What where the commands/arguments used for compiling libs for these projects by you, the developers?
If all works out I would also be willing to post a PR which includes updates to the README pages for anyone who tries to do the same in the future.

Sokkero:
Opening this PR to document my findings and close #546. Adding a script for building a static lib for ios, some minor changes to the build_android.sh script to make usage of it clearer and adding to the build documentation.

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -1,11 +1,37 @@
 ## Build
 
+### Building static libraries
+
+A static library can be build using:
+```
+mkdir build #make a build dir so that you can build out of tree
+cd build
+cmake -DUSE_TLS=1 ..
+make -j
+```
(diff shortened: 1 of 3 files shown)


bsergean (approved):

bsergean:
thanks !

PR #550, opened 2025-05-09
IXWebSocketTransport: Avoid bloating _rxbuf

### LINKED ISSUE #429: O(N^2) time complexity bug in WebsocketTransport
https://github.com/machinezone/IXWebSocket/blob/679ce519dd0d6d3919990abdee7109a5eeb99aa0/ixwebsocket/IXWebSocketTransport.cpp#L1067
https://github.com/machinezone/IXWebSocket/blob/679ce519dd0d6d3919990abdee7109a5eeb99aa0/ixwebsocket/IXWebSocketTransport.cpp#L691

The way the code is using std::vector for the receive buffer with data being added at the end and removed from the front creates a time complexity bug. When erasing at the front, vector::erase is O(N) where N is the size of the whole vec, not the size of the packet data being erased. That means that dequeuing all packets is actually O(N^2)

If for whatever reason the receive buffer grows big enough, the time to dequeue one packet can exceed the time to receive one packet at which point it's a downward spiral that cannot be recovered (the process will grind to a halt). 

This is not just theoretical, I experienced it in production.

The issue can be fixed by using a different datastructure for the receive buffer, such as std::deque.

awelzel:
Buffering too much data into _rxbuf results in dispatch() processing to run into a performance issue due to calling _rxbuf.erase() for each frame.

Concretely, if _rxbuf grows large due to a client sending frames very fast, each processing in dispatch() results in moving the remaining buffer to the front.

Instead of restructuring to avoid .erase(), this patch limits the maximum size of _rxbuf to kChunkSize to alleviate the O^2 erase() overhead. It also has the side-effect of keeping the received data in the OS's TCP stack, building up back pressure to the client earlier if the server code can't keep up.

Fixes #429

### DIFF
--- a/ixwebsocket/IXWebSocketTransport.cpp
+++ b/ixwebsocket/IXWebSocketTransport.cpp
@@ -547,8 +547,11 @@ namespace ix
 
             if (_rxbuf.size() < ws.header_size + ws.N)
             {
+                _rxbufWanted = ws.header_size + ws.N;
                 return; /* Need: ws.header_size+ws.N - _rxbuf.size() */
             }
+
+            _rxbufWanted = 0;
 
             if (!ws.fin && (ws.opcode == wsheader_type::PING || ws.opcode == wsheader_type::PONG ||
                             ws.opcode == wsheader_type::CLOSE))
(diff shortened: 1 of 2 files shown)


awelzel:
With this patch, the reproducer code from #429 completes in 3 to 4 seconds.
```
 /usr/bin/time python3 client-sender.py --url ws://localhost:1234 --messages 500000
0 Hello!
1.57user 1.77system 0:03.36elapsed 99%CPU (0avgtext+0avgdata 24832maxresident)k
0inputs+0outputs (0major+3880minor)pagefaults 0swaps
```

Without the patch, I've observed it to take up to 60 seconds locally, but depends on how the buffer builds up over time.

I do think the 32k kChunkSize that is read by default is a bit large, but would rather not change too many things.

awelzel:
The new server-side flamegraph shows that the `.erase()` overhead is essentially gone now. There's some overhead visible copying from `_rxbuf` into the `frameData` string. Not attempting to fix this here and mostly mentioning that as a future improvement could be to read client side data directly into a `frameData` buffer instead of using an intermediary `_rxbuf` in order to avoid the extra copy.

stacks-fix

PR #552, opened 2025-05-22
Do not invoke message callbacks if we aren't connected.

bsergean:


bsergean:
Other option is to not fail send() the same way when we aren't connected, taking out that early out.

I haven't run the autobahn test for a while btw ... I wonder if those fixes break something or not spec wise

PR #564, opened 2026-01-06
Update IXWebSocketHttpHeaders.cpp

zjp8369:
修改header中值为空时，程序崩溃问题

### DIFF
--- a/ixwebsocket/IXWebSocketHttpHeaders.cpp
+++ b/ixwebsocket/IXWebSocketHttpHeaders.cpp
@@ -12,6 +12,23 @@
 
 namespace ix
 {
+     static inline void ltrim(std::string& s) {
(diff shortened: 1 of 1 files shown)


bsergean:
Thanks for identifying this bug! I've opened #574 as an alternative fix that addresses the same root cause with a few improvements:

- Adds the missing bounds check on the space-skipping `while` loop (UB if colon is at the end of the line)
- Replaces the `size - start - 2` unsigned arithmetic with a trailing trim — avoids `size_t` underflow and handles both `\r\n` and `\n`-only line endings
- Fixes the `isspace` UB by casting to `unsigned char` in the lambda
- Removes the redundant 1024-byte cap (the buffer is already capped at 1023 chars)

PR #569, opened 2026-03-01
Add support for LibreSSL

SigmaTel71:
The frontends are mostly interchangeable between OpenSSL and LibreSSL.

I thought it would be useful for a project licensed under BSD license to support a cryptography library from its "homeland" :)
As the first line says, LibreSSL is not binary compatible with OpenSSL but provides the same interfaces for applications that use the latter so these could be recompiled with the former.
The diff to looks a bit large to me and I was thinking there should be no intentions to touch preprocessor conditions and implicitly treat LibreSSL as OpenSSL.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,16 +125,16 @@ option(USE_TLS "Enable TLS support" FALSE)
 if (USE_TLS)
     # default to securetranport on Apple if nothing is configured
     if (APPLE)
-      if (NOT USE_MBED_TLS AND NOT USE_OPEN_SSL) # unless we want something else
+      if (NOT USE_MBED_TLS AND NOT USE_OPEN_SSL AND NOT USE_LIBRE_SSL) # unless we want something else
         set(USE_SECURE_TRANSPORT ON)
       endif()
     # default to mbedtls on windows if nothing is configured
     elseif (WIN32)
-      if (NOT USE_OPEN_SSL) # unless we want something else
+      if (NOT USE_OPEN_SSL AND NOT USE_LIBRE_SSL) # unless we want something else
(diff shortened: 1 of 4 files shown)


bsergean:
Sorry I lost write permission to this repo. (I am the owner).
We need to contact tripledots studio but I cannot get a hold of them, if you can help with that it would be excellent (twitter ? contact email ?).

Thanks for the PR would love to merge it.

SigmaTel71:
> We need to contact tripledots studio but I cannot get a hold of them, if you can help with that it would be excellent (twitter ? contact email ?).

Can't find one on their website and the public email that Machine Zone has is for press inquiries. Or should I try the open-source Machine Zone email instead?

andymillermz:
Thanks, @SigmaTel71 and @bsergean!

@bsergean We're looking into this - I'll see if we can't get you back merge permissions again 👍

bsergean:
Great thanks @andymillermz !

SigmaTel71:
> Great thanks @andymillermz !

As things are taking off the ground, what do you think of the current diff? Is there anything that I missed or should implement besides the option to compile with a different backend?

bsergean:
Looked at the code change and it looks good to me @SigmaTel71

bsergean (approved):

SigmaTel71:
> Looked at the code change and it looks good to me @SigmaTel71

I converted the PR from draft to ready to merge if the code is okay.

andymillermz:
@bsergean I've granted you permissions on this repo again! Thanks for all the work here!

SigmaTel71:
Pushed a minor QoL fix, LibreSSL version is printed during configuration now, as it would with OpenSSL.

PR #571, opened 2026-04-18
IPv6 bracketed address not functional when used on client side Linux/macOS

cosminpolifronie:
Hello! In a project at work we are running into trouble using IXWebSocket with bracketed IPv6 addresses, specifically [::1]:1234, and specifically on Linux. Windows worked fine, somehow.

This pull request fixes 2 issues:
1. Bracketed IPv6 addresses not being properly parsed by clParseURL::ParseURL (commit 1)
2. Even after proper parsing, client was not able to connect to server on Linux/macOS because of RFC 7230 violation (brackets are required in Host and Origin headers) (commit 2)

These issues have been manually debugged and fixed at first. I wasn't a fan of the `clParseURL::ParseURL` readability, and my manual fix for the second issue became a bit too complex for comfort in a codebase I'm not familiar with.

As such, I wanted to see how Copilot would handle this + refactoring the `clParseURL::ParseURL` logic and error handling.

This isn't AI slop. All the changes were human reviewed every step of the way and edge case handling was insisted on (I did find some that Copilot didn't handle). Also, tests were introduced for all the modified behavior.

All tests are passing on macOS. They have been also run with fixes disabled, and expectedly failed.

TL;DR: the code and tests are Copilot written but have been human prompted and reviewed. Initial investigation and fix was done 100% manually.
TODO: check the implementation on Linux and Windows (checked only macOS for now), and in our specific project (will come back on Monday with results)

Great project, by the way. This library is a breeze to use compared to Boost.Beast. Your work is much appreciated!

PR #572, opened 2026-04-25
Fix build warnings

cosminpolifronie:
There were lots of build warnings on MacOS because of using lots of old dependencies.
All of them have been updated to the latest versions and removed from the repository.
All dependencies are now downloaded from GitHub, using the latest tags or commits at the time of writing.

The only remaining MacOS warning is given in AppleSSL, as SecureTransport has been deprecated since macOS 10.15.
I am not familiar with those TLS paths and there are no TLS tests in the library so I don't think I'm the right person to add a Network.framework implementation, even if AI would implement it.

I will leave that to someone who has an interest in using the newer Apple SSL framework.


bsergean:
There are too many changes bundled into this.

Could you break it down into a few different ones, one of them being using cmake to fetch dependencies. The build is failing btw for some reason.

cosminpolifronie:
The commits can be reviewed one by one, the changes are not big for each (they are quite small for a pull request each) and are exactly described in the commit messages.

I will take another look at the builds. It seems the GH builds you have apply much stricter warning flags and error on warning.

It would have been nice to have those included in the root CMakeLists.txt.

cosminpolifronie:
I can see the builds were started before #571 was merged, which contained some mbedtls build system fixes. Could you please re-run the builds so I can focus on the actual issues introduced by this PR?

cosminpolifronie:
It seems there were some manual changes done to msgpack11 that fixed the exact build failures we are seeing here.
I have pushed those changes upstream.
The builds should pass after the PR is merged upstream and IXWebSocket starts using it.
https://github.com/ar90n/msgpack11/pull/19

cosminpolifronie:
@bsergean Hi! There is no answer from upstream, I wouldn't wait for the cleanup in this PR.

Could you please create a release tag with what is currently available on develop? We would like to use the fixes in production via conan2 dependency.

bsergean:
@cosminpolifronie I just made a release !

cosminpolifronie:
@bsergean This is ready for review and merge.

bsergean:
thanks !

PR #582, opened 2026-06-06
Fix include dir & version

xiaozhuai:
See also https://github.com/microsoft/vcpkg/pull/52198#pullrequestreview-4440489536


BillyONeal (approved):

BillyONeal:
Repeating the content behind the link. @xiaozhuai filed a PR to update the vcpkg port and in that PR, GPT 5.4 found the following. (It did not write anything in this response)

The installed .pc file still claims that this is 11.4.6 despite this claiming to be 12.0.0. https://github.com/machinezone/IXWebSocket/blob/998cf95dc7f8cf182c54634cc1c2787abc6cb6be/CMakeLists.txt#L9

The installed CMake and pkgconfig bindings appear to be inconsistent. `share\ixwebsocket\ixwebsocket-targets.cmake` has:

```cmake
set_target_properties(ixwebsocket::ixwebsocket PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "IXWEBSOCKET_USE_TLS;IXWEBSOCKET_USE_MBED_TLS;IXWEBSOCKET_USE_ZLIB"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/ixwebsocket"
```

but `lib\pkgconfig\ixwebsocket.pc` has:

```
prefix=${pcfiledir}/../..
includedir=${prefix}/include
Cflags: "-I${includedir}"
```

This means that the CMake bindings want the user to say, for example, `#include <IXUuid.h>`, but the pkgconfig bindings want the user to say `#include <ixwebsocket/IXUuid.h>`.

https://github.com/machinezone/IXWebSocket/blob/998cf95dc7f8cf182c54634cc1c2787abc6cb6be/test/IXUrlParserTest.cpp#L10 looks like the intent is for the .pc to be correct. This PR fixes both of those problems (though likely you will want 12.0.1 or something so that you don't mutate an existing "release")

bsergean:
Thanks @BillyONeal will try to get that fixed today. Changing the cmake include path worries me a bit.

PR #585, opened 2026-06-12
Fix double-close of _serverFd in SocketServer teardown

zardus:
Hi, I'm using IXWebSocket in https://github.com/openglad/openglad, and ran into an issue when using this in a multithreaded setting.

It seems that `SocketServer::stop()` closes `_serverFd` but never clears it, and stop() runs again from both `~WebSocketServer()` and `~SocketServer()`, so every teardown closes the same fd number two or three times. If another thread opens a descriptor between those closes, the kernel reuses the number and the stale close destroys it. I was hitting intermittent process aborts ("Unexpected error 9 on netlink descriptor", glibc's getaddrinfo netlink probe) whenever a DNS lookup in another thread raced a server teardown.

This PR sets `_serverFd` to -1 after closing it (including in error paths), like what seems to happen to `_sockfd` currently elsewhere in the code. This seems to have fixed it for me.

### DIFF
--- a/ixwebsocket/IXSocketServer.cpp
+++ b/ixwebsocket/IXSocketServer.cpp
@@ -95,6 +95,7 @@ namespace ix
                << "at address " << _host << ":" << _port << " : " << strerror(Socket::getErrno());
 
             Socket::closeSocket(_serverFd);
+            _serverFd = -1;
             return std::make_pair(false, ss.str());
         }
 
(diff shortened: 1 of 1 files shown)


bsergean:
Do you think you can merge master in your branch, I think I fixed the build issue. with uwp, then we'll reckick the PR checks.

zardus:
Done!

PR #595, opened 2026-07-15
Make server-side pre-connection errors observable via a log callback

### LINKED ISSUE #593: WebSocketServer (server-side TLS failures) produce zero application-visible logs while WebSocket client reports them via Error callback
### Problem

When a TLS handshake fails before a WebSocket connection is established, the client reports it via `setOnMessageCallback` (`WebSocketMessageType::Error`), so applications can log it normally.

On the server, the same class of failure is not delivered to `setOnClientMessageCallback`. The library only calls internal `logError()`, which writes to stderr. In our deployment stderr is not part of the main application log stream, so operators see nothing useful on the server side when TLS fails.

### Reproduction

1. Run a `WebSocketServer` with TLS enabled and `setOnClientMessageCallback` wired to the app logger.
2. Trigger a TLS failure (e.g. invalid CA, cert mismatch, bad client cert).
3. Client logs show TLS/connect errors via the Error callback.
4. Server application logs show nothing — no error, no rejected handshake, no client IP.

### Expected

Server-side TLS accept/handshake failures should be observable by application code (callback or pluggable logger), similar to the client Error callback.

### Actual

Failures are only emitted internally to stderr; the application callback is never invoked for pre-connection TLS errors.

### Environment

- IXWebSocket version/commit: `<fill in>`
- TLS backend: OpenSSL
- OS: Linux

bsergean:
Fixes #593

## Problem

When a TLS handshake fails on the server before a WebSocket connection is established, the error is only written to stderr via the internal `logError()`. Applications wired to `setOnClientMessageCallback` see nothing — no error, no client IP — while the client side reports the same class of failure through its `Error` message callback. The handshake runs in `SocketServer::run()` before any `WebSocket` object exists, so the existing callbacks can't be reached at that point.

## Changes

- **`SocketServer::setLogCallback(LogLevel level, const std::string& message)`** — when set, server log messages (errors and info) are delivered to the callback instead of being written to stderr/stdout. This covers all pre-connection failures: TLS accept, socket accept errors, max-connections rejections, socket creation. Works for `WebSocketServer` and `HttpServer` alike. Default behavior is unchanged when no callback is set.
- The `tls accept failed` and `cannot create socket` error messages now include the client `ip:port`, which was already known at that point but not reported.
- New unittest: starts a `wss` server, connects with a plaintext client so the TLS handshake fails, and asserts the error (including the peer address) reaches the log callback.
- Documented in `docs/usage.md`.

Example log line now delivered to the application:

```
SocketServer::run() tls accept failed for client 127.0.0.1:59898: OpenSSL failed - error:0A00009C:SSL routines::http request
```

## Test plan

- `IXWebSocketServerTest` passes locally (macOS, OpenSSL build), including the new TLS-failure test.
- Full ninja build (library, tests, ws) clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -457,6 +457,25 @@ int pingIntervalSeconds = 45;
 ix::WebSocketServer server(port, host, backlog, maxConnections, handshakeTimeoutSecs, addressFamily, pingIntervalSeconds);
 ```
 
+### Server log callback
+
+By default the server writes internal errors to stderr (and some info messages to stdout). Errors that happen before a connection is fully established — for example a failed TLS handshake when a client presents a bad certificate — cannot be reported through `setOnClientMessageCallback`, since no WebSocket object exists yet at that point.
+
+To capture those messages in your application logs, set a log callback. It applies to `WebSocketServer` as well as `HttpServer`. When a callback is set, messages are delivered to it instead of being written to stderr/stdout.
+
+```cpp
+server.setLogCallback([](ix::LogLevel level, const std::string& msg) {
+    if (level == ix::LogLevel::Error)
+    {
+        myLogger.error(msg);
+    }
+    else
+    {
+        myLogger.info(msg);
+    }
+});
+```
+
 ## HTTP client API
 
 ```cpp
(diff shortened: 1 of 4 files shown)


shivaxsaib:
@bsergean Hey thanks for the swish update on this. Can you please tell when will be the next release?

PR #604, opened 2026-08-17
Support FD_CLOEXEC for server and client sockets created by SocketServer

awelzel:
This adds a method setCloseOnExec() to enable setting FD_CLOEXEC on the server and client file descriptors before starting to listen or initiating the WebSocket handshake. This is useful if a process spawns child processes via system() or otherwise, but does not want have WebSocket specific sockets to be inherited to child processes.


bsergean:
Thanks !

PR #607, opened 2026-08-28
Fix: Set soversion

silverducks143:
Greetings!

Currently, soversion is unset, and thus resulting library defaults to full version with patch as it's soname.
This PR adds soversion definition to `CMakeLists.txt`, equal to major version.


bsergean:
Thanks @silverducks143 !

